PR #16, opened 2020-02-19
feat: check table and attribute names

fdlk:


sidohaakma:
Close because of this one: https://github.com/molgenis/molgenis-service-datashield/pull/18

PR #21, opened 2020-02-25
feat: list installed packages

fdlk:


src/main/java/org/molgenis/datashield/DataController.java
```diff
@@ -62,4 +74,31 @@ public void load(@PathVariable String entityTypeId)
...
     return datashieldSession.execute(connection -> executorService.execute(cmd, connection));
   }
+
+  @GetMapping(value = "/packages")
+  public List<Map<String, String>> getPackages() throws REXPMismatchException, RserveException {
+    return datashieldSession.execute(
+        connection -> toStringMap(connection.eval(INSTALLED_PACKAGES)));
+  }
+
+  private static List<Map<String, String>> toStringMap(REXP matrix) throws REXPMismatchException {
```
fdlk: very clumsily transforms matrix of package info to something we can brew json of

PR #22, opened 2020-02-27
feat: add formatter

fdlk:


src/main/java/org/molgenis/datashield/r/Formatter.java
```diff
@@ -0,0 +1,41 @@
...
+          .put("\u000B", "\\v")
+          .build();
+
+  public static final CharSequenceTranslator ESCAPE_R =
+      new AggregateTranslator(
+          new LookupTranslator(Map.of("\"", "\\\"", "\\", "\\\\")),
+          new LookupTranslator(R_CTRL_CHARS_ESCAPE),
+          UnicodeEscaper.outsideOf(32, 0x7f));
+
+  public static String quote(final CharSequence input) {
```
fdlk: It's recommended to use `"` in R strings, not `'`. It does make string literals with expected values a bit less nice in tests, but otoh the executed commands will be more standard.

PR #31, opened 2020-03-06
Add /exists endpoint

tommydeboer:


datashield/src/main/java/org/molgenis/datashield/DataController.java
```diff
@@ -2,7 +2,10 @@
 
 import static org.molgenis.datashield.DataShieldUtils.createRawResponse;
 import static org.molgenis.datashield.DataShieldUtils.serializeCommand;
-import static org.springframework.http.MediaType.*;
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
```
sidohaakma: Is this better than the start-import. Just out of curiosity.

sidohaakma: Ah, I get it. It is more explicit.

PR #34, opened 2020-03-18
Use Datashield4J

tommydeboer:


datashield/src/main/java/org/molgenis/datashield/service/DataShieldExpressionRewriterImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    this.aggregateEnvironment = new DatashieldEnvironment(DSMethodType.AGGREGATE);
+    this.assignEnvironment = new DatashieldEnvironment(DSMethodType.ASSIGN);
+
+    this.rAggregateScriptGenerator = new RScriptGenerator(aggregateEnvironment);
+    this.rAssignScriptGenerator = new RScriptGenerator(assignEnvironment);
+  }
+
+  @PostConstruct
+  public void populateEnvironments() throws RserveException, REXPMismatchException {
+    RConnection connection = rConnectionFactory.retryCreateConnection();
```
fdlk, line 53: try/finally close the connection

datashield/src/main/java/org/molgenis/datashield/service/DataShieldExpressionRewriterImpl.java
```diff
@@ -0,0 +1,114 @@
...
+    List<Package> packages = packageService.getInstalledPackages(connection);
+
+    packages.stream()
+        .map(Package::aggregateMethods)
+        .filter(Objects::nonNull)
+        .flatMap(Set::stream)
+        .map(this::toDSMethod)
+        .forEach(m -> addToEnvironment(m, aggregateEnvironment));
```
fdlk, line 61: Should we detect collisions, i.e. if multiple packages declare the same method differently?

datashield/src/main/java/org/molgenis/datashield/service/DataShieldExpressionRewriterImpl.java
```diff
@@ -0,0 +1,114 @@
...
+  private final PackageService packageService;
+  private final RConnectionFactory rConnectionFactory;
+
+  private final DSEnvironment aggregateEnvironment;
+  private final DSEnvironment assignEnvironment;
+
+  private final RScriptGenerator rAggregateScriptGenerator;
+  private final RScriptGenerator rAssignScriptGenerator;
+
+  public DataShieldExpressionRewriterImpl(
```
fdlk, line 39: This class does a bit much, both keeping track of the methods and wiring all the parts together.

datashield/src/main/java/org/molgenis/datashield/service/DataShieldExpressionRewriterImpl.java
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  /**
+   * Method strings come in two forms: either without a package ('meanDS'), meaning they belong to
+   * the 'dsBase' package, or with a name and a package ('dim=base::dim'), meaning they are part of
+   * another package instead of 'dsBase'.
+   */
+  private PackagedFunctionDSMethod toDSMethod(String method) {
+    if (method.contains("=")) {
+      String[] nonDsBaseMethod = method.split("=");
```
fdlk, line 78: make sure the array has exactly two elements

datashield/src/main/java/org/molgenis/datashield/DataController.java
```diff
@@ -89,8 +93,9 @@ public void load(@PathVariable String entityTypeId, @PathVariable String assignS
 
   @PostMapping(value = "/execute", consumes = TEXT_PLAIN_VALUE, produces = APPLICATION_JSON_VALUE)
   public CompletableFuture<ResponseEntity<Object>> execute(
-      @RequestBody String cmd, @RequestParam(defaultValue = "false") boolean async) {
-    CompletableFuture<REXP> result = datashieldSession.schedule(cmd);
+      @RequestBody String expression, @RequestParam(defaultValue = "false") boolean async) {
+    String rewrittenExpression = expressionRewriter.rewriteAggregate(expression);
```
fdlk, line 97: We'll need to handle the exception so we don't throw a 500 when the expression is incorrect

datashield/src/main/java/org/molgenis/datashield/exceptions/DataShieldExpressionException.java
```diff
@@ -3,7 +3,10 @@
 import static java.lang.String.format;
 
 import org.obiba.datashield.r.expr.ParseException;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseStatus;
 
+@ResponseStatus(HttpStatus.BAD_REQUEST)
```
fdlk, line 9: \o/

PR #37, opened 2020-03-23
demo: web socket interaction R <-> Server

fdlk:


sidohaakma:
We choose to implement the async communication flow.

PR #40, opened 2020-03-30
feat: implement last command

sidohaakma:


datashield/src/main/java/org/molgenis/datashield/pojo/DataShieldCommand.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+
+public class DataShieldCommand<T> {
+  private final Instant createDate;
+  private final UUID id;
+  private final String expression;
+  private final boolean withResult;
+  @SuppressWarnings("java:S3077") // CompletableFuture is thread-safe
```
sidohaakma: Nice, these are the properties that need volatile I pressume.

fdlk: Yes, I think so, the ones that are mutable should be synchronized across threads

PR #49, opened 2020-04-21
feat: user workspaces

fdlk:
TODO: 
- [x] Make separate buckets for shared workspaces and user workspaces
- [x] Sanitize workspace IDs with some regex (Opal implementation allows everything: https://github.com/obiba/opal/search?q=saveId&unscoped_q=saveId) WIP
- [x] Change endpoints for actions to request parameter instead of pathvariable
- [x] Filter out the table env when saving workspace
- [ ] What with files on harddisk?


datashield/test-workspaces.http
```diff
@@ -0,0 +1,26 @@
...
+DELETE http://localhost:8080/symbols/D
+
+### List user workspaces
+GET http://localhost:8080/workspaces
+
+### Create a user workspace
+POST http://localhost:8080/workspaces/test
+
+### Load a user workspace
+POST http://localhost:8080/load-workspace/test
```
fdlk, line 23: Why not load-workspace?workspace=test ?
And then same for load-tables?workspace=X&workspace=Y or load-tables?workspaces=X,Y

PR #51, opened 2020-04-23
chore: add adr concerning RData

sidohaakma:


doc/adr/0006-use-rdata-format-as-data-input.md
```diff
@@ -0,0 +1,23 @@
...
+
+We want to make the MOLGENIS "Armadillo" service data provider agnostic. There are a couple of reasons why we are doing this
+- the service is usable for other parties as well
+- the service can still integrate with MOLGENIS. 
+- the release cycle of the service is data provider independent
+- the service can be developed by other parties as well
+
+## Decision
+
+We implement and endpoint to upload and load RData files in the MOLGENIS "Armadillo" service to manage data for the use in DataSHIELD.
```
mswertz, line 19: and => an

doc/adr/0006-use-rdata-format-as-data-input.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Consequences
+- We need a client which is capable to deal with uploading RData files into the MOLGENIS "Armadillo" service
+- MOLGENIS needs to have an RData download option to be able to communicate directly with the MOLGENIS "Armadillo" service
```
mswertz, line 23: or EMX => RData convertor

doc/adr/0007-use-roles-to-authorise-users-to-use-data.md
```diff
@@ -0,0 +1,39 @@
...
+
+**ID-based**
+The ID-based approach was suggested as a first solution to make the service secure. A big disadvantage is that anyone with the link can 
+share the data. That means that there has to be a lot of trust between the data manager and the researcher. In practices, this is not feasible.
+
+## Decision
+
+We will use roles as a basic principle to give users permission on the data they need to have access to. 
+Within the MOLGENIS "Armadillo" service there are at least these 2 types of roles:
+- Researcher
```
mswertz, line 26: I expect that in time we will need 'role per study'. Do you agree? 
So how will that play out? Is that then implemented by 'read' on a particular folder?

doc/adr/0008-load-multiple-rdata-files.md
```diff
@@ -0,0 +1,18 @@
+# 8. As a researcher you are able to load multiple RData files
```
mswertz, line 1: Is that within one study? Or all studies the researcher has access too?

doc/adr/0007-use-roles-to-authorise-users-to-use-data.md
sidohaakma: That is correct. It should match the role descriptions below, but indeed you can have for instance 10 study folders with subsets of the data in it and apply read rights for different researchers at a folder level.

doc/adr/0008-load-multiple-rdata-files.md
sidohaakma: If you want rights on a specific file in a folder it is not possible. Maybe I should write that down in here?

fdlk: If a researcher participates in multiple studies, he can load data files from all studies. (But the tables are not explicitly namespaced, if different .RData files have different versions of the same table only the last one loaded will survive.)

PR #52, opened 2020-04-24
feat: info endpoint

fdlk:
Uses spring boot actuator's info endpoint to give server info to the client.﻿


sidohaakma:
/actuator/health works as well ;-D.

PR #55, opened 2020-04-28
feat: implement /methods endpoints

fdlk:


tommydeboer:
It wasn't immediately clear how to use the parameter and I think the error message could be a bit more helpful:

```
"Failed to convert value of type 'java.lang.String' to required type 
'org.obiba.datashield.core.DSMethodType'; nested exception is 
org.springframework.core.convert.ConversionFailedException: Failed to convert from type 
[java.lang.String] to type [@org.springframework.web.bind.annotation.RequestParam 
org.obiba.datashield.core.DSMethodType] for value 'dd'; nested exception is java.lang.IllegalArgumentException: No enum constant org.obiba.datashield.core.DSMethodType.dd"
```

Maybe return a list of possible values?

fdlk:
The message is the one generated by spring when you send invalid enum value. I considered converting manually but I don’t really expect lots of folks to be calling it manually, just the client.

PR #88, opened 2020-08-06
chore: by default do not debug RServe

fdlk:


armadillo/docker-compose.yml
```diff
@@ -25,8 +25,8 @@ services:
       - 3000:8080
   rserver:
     image: molgenis/ds-60-rserver-363:latest
-    environment:
-      DEBUG: "TRUE"
+#    environment:
+#      DEBUG: "TRUE"
```
tommydeboer, line 29: Why not say `DEBUG = "FALSE"` instead?

PR #95, opened 2020-08-13
fix: added contenttype to non-async flow and added release flow in ci

sidohaakma:


sidohaakma:
[add debug info to all datashield service methods](https://trello.com/c/CqRiyfyh/374-add-debug-info-to-all-datashield-service-methods)

sidohaakma:
[dsBaseClient errors](https://trello.com/c/hFXNujz5/373-dsbaseclient-errors)

sidohaakma:
[added release flow to the armadillo](https://trello.com/c/aTenTyKP/325-added-release-flow-to-the-armadillo)

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloUtils.java
```diff
@@ -22,10 +27,13 @@ public static String serializeExpression(String cmd) {
     byte[] rawResult = new byte[0];
     if (result.isRaw()) {
       try {
+        LOGGER.trace(format("RAW result : [ %s ]", new String(result.asBytes(), UTF_8)));
         rawResult = result.asBytes();
       } catch (REXPMismatchException e) {
         throw new IllegalStateException(e);
       }
+    } else {
```
fdlk, line 35: Let's just get really upset and throw exception if it's not a raw result. Fail fast, fail hard.

armadillo/src/main/java/org/molgenis/armadillo/DataController.java
```diff
@@ -183,7 +183,10 @@ public void removeSymbol(@Valid @Pattern(regexp = SYMBOL_RE) @PathVariable Strin
   }
 
   @Operation(summary = "Execute expression")
-  @PostMapping(value = "/execute", consumes = TEXT_PLAIN_VALUE)
+  @PostMapping(
+      value = "/execute",
+      consumes = TEXT_PLAIN_VALUE,
+      produces = APPLICATION_OCTET_STREAM_VALUE)
```
fdlk, line 189: nice catch!

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloUtils.java
```diff
@@ -22,10 +27,13 @@ public static String serializeExpression(String cmd) {
     byte[] rawResult = new byte[0];
     if (result.isRaw()) {
       try {
+        LOGGER.trace(format("RAW result : [ %s ]", new String(result.asBytes(), UTF_8)));
```
fdlk, line 30: check LOGGER.isTraceEnabled() to save some time (sonar warning)

Jenkinsfile
```diff
@@ -118,13 +118,46 @@ pipeline {
...
             }
         }
-    }
-    post {
-        success {
-            hubotSend(message: 'Build success', status: 'INFO', site: 'slack-pr-app-team')
-        }
-        failure {
-            hubotSend(message: 'Build failed', status: 'ERROR', site: 'slack-pr-app-team')
+        stage('Steps: [ x.x ]') {
```
fdlk, line 121: Not completely sure about the x.x approach as opposed to only releasing from master. I would hope the service is simple enough that we can simply steam forward in a straight line ahead without the need for backports and stable branches.

fdlk: But am happy to go ahead with this if it's easiest and/or you think it's the smart thing to do

PR #97, opened 2020-08-17
fix: buffer R file outputstream

fdlk:
The RFileOutputStream that we use is not buffered, after experimenting some it looks like adding a buffer will speed up performance quite significantly.

transfer
Transfer times for different buffer sizes of the copy buffer (red) and output stream buffer (blue)


tommydeboer:
Nice!

sidohaakma (approved):

PR #98, opened 2020-08-18
feat: kill r processes

fdlk:


armadillo/src/main/java/org/molgenis/armadillo/ArmadilloSession.java
```diff
@@ -61,20 +66,31 @@ void tryDetachRConnection(RConnection connection) {
...
-      return rSession.attach();
+      var connection = rSession.attach();
+      rSession = null;
+      return connection;
     } catch (RserveException err) {
       throw new ArmadilloSessionException("Could not attach connection to RSession", err);
     }
   }
 
-  public synchronized void sessionCleanup() {
```
fdlk, line 72: No longer needs to wait for evaluation to be finished, it kills the process instead!

sidohaakma:
[Cleanup stale sessions on the Rserve environment](https://trello.com/c/Y3NuiRsk/214-cleanup-stale-sessions-on-the-rserve-environment)

r/src/main/java/org/molgenis/r/service/ProcessServiceImpl.java
```diff
@@ -41,36 +43,57 @@
...
+    this.rExecutorService = rExecutorService;
   }
 
   @Override
   public int countRserveProcesses(RConnection connection) {
     try {
-      return ((REXP) connection.eval(COUNT_RSERVE_PROCESSES_COMMAND).asList().get(0)).asInteger();
-    } catch (RserveException | REXPMismatchException | ClassCastException e) {
+      return ((REXP)
+              rExecutorService.execute(COUNT_RSERVE_PROCESSES_COMMAND, connection).asList().get(0))
```
sidohaakma, line 65: Nice to use the RExecutorService here!

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloSession.java
```diff
@@ -61,20 +66,31 @@ void tryDetachRConnection(RConnection connection) {
   RConnection getRConnection() {
     try {
       if (rSession == null) {
-        return connectionFactory.createConnection();
+        var connection = connectionFactory.createConnection();
+        pid = processService.getPid(connection);
+        return connection;
```
sidohaakma, line 71: Maybe log the pid as well? We ports, but pid's would be useful as well concerning you can execute a api call to kill it?
Or do we use the connection to determine the pid using the actuator endpoints?

sidohaakma (changes requested):

r/src/main/java/org/molgenis/r/service/ProcessServiceImpl.java
fdlk: Yes, I'd forgotten all about it :D

sidohaakma (approved):

PR #106, opened 2020-09-21
feat: retrieve package info from inst/DATASHIELD

fdlk:


r/src/main/java/org/molgenis/r/service/PackageServiceImpl.java
```diff
@@ -34,9 +31,26 @@
...
+          + "  file <- system.file(\"DATASHIELD\", package = package$Package)\n"
+          + "  if (file == \"\") package\n"
+          + "  else {\n"
+          + "    from_file <- as.list(unlist(tibble::as_tibble(read.dcf(file))))\n"
+          + "    result <- append(as.list(package), as.list(from_file))\n"
+          + "    result[!is.na(result)]\n"
+          + "  }\n"
+          + "}\n"
+          + "\n"
+          + "installed.packages(fields = c('AssignMethods', 'AggregateMethods', 'Options')) %>%\n"
```
sidohaakma, line 50: You do not want to use the constants here?

sidohaakma (approved):
I have added the resourcer package to the whitelist. I have done this because we published the docker already, it breaks when you not add the resourcer package.

PR #111, opened 2020-10-01
chore: make it easier to add dsDanger in dev environment

fdlk:
Based on #110


docker/rserver/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM molgenis/rserver:latest
+RUN install2.r remotes
+RUN installGithub.r datashield/dsDanger@v6.1-dev datashield/dsBase@v6.1-dev
```
sidohaakma, line 3: Maybe add the versions as build-args?

sidohaakma (changes requested):

sidohaakma (approved):

PR #133, opened 2021-07-14
feat: upgrade datashield4j to 2.0.0

sidohaakma:
Upgraded the datashield4j libraries in the armadillo service to version 2.0.0
- **Infrastructure**
  - Changed the location of the repository of the libraries to https://obiba.jfrog.io/artifactory/libs-release-local/ instead of bintray
  - Configure the nexus to use the new repository as well
- **Code**
   - Used new parser in the code. There is a new R-parser (version 2) which is less lenient then old parser (version 1)
  - The way you now create the assign and aggregate parsers is different
  - The old parser still uses  `PackagedFunctionDSMethod`. The new parser uses `DefaultDSMethod`

The exact behavior we can not describe here because of the nature of the bug.


sidohaakma:
[Security update for DataSHIELD](https://trello.com/c/x23ozuzY/168-security-update-for-datashield)

fdlk (approved):

PR #145, opened 2022-01-21
fix: seed option is not set correctly

sidohaakma:


sidohaakma:
[Fix seed option](https://trello.com/c/XDS9676G/64-fix-seed-option)

tommydeboer (approved):

PR #148, opened 2022-02-15
feat: load whitelisted packages only

sidohaakma:


armadillo/src/main/resources/application.yml
```diff
@@ -18,11 +21,16 @@ datashield:
...
           seed: 342325352
 #          foo: bar
-    - name: exposome
-      environment: exposome
+#    - name: exposome
+#      environment: exposome
+#      whitelist:
+#        - dsBase
+#        - dsExposome
+    - name: omics
```
tommydeboer, line 29: Should we comment out the `omics` profile too?

r/src/main/java/org/molgenis/r/service/PackageServiceImpl.java
```diff
@@ -68,6 +72,18 @@ public PackageServiceImpl(REXPParser rexpParser) {
     }
   }
 
+  @Override
+  public void loadPackages(RConnection connection, Set<String> pkgs) {
+    try {
+      String packages = String.format("\"%s\"", String.join("\",\"", pkgs));
+      LOGGER.debug("Loading packages [ {} ]", packages);
+      connection.eval("library(" + packages + ")");
+      LOGGER.debug("Successfully loaded packages [ {} ]", packages);
```
tommydeboer, line 81: Maybe demote this one to `trace`

tommydeboer: Or remove it

tommydeboer (changes requested):

armadillo/src/main/resources/application.yml
sidohaakma: Yes, it needs to be done in de docker-compose as well. I will do this.

r/src/main/java/org/molgenis/r/service/PackageServiceImpl.java
sidohaakma: I will demote it to trace.

README.md
```diff
@@ -30,3 +30,20 @@ To spin up your own server on a laptop, you can run `docker-compose up` in the d
 
 Alternatively for [kubernetes](https://k8s.io) deployments look at the [helm chart](https://github.com/molgenis/molgenis-ops-helm/tree/master/charts/molgenis-armadillo). 
 Or for virtual server deployment look at the [ansible playbook](https://galaxy.ansible.com/molgenis/armadillo).
+
+### Development
+You need to start several backend services in other to be able to develop in the Armadillo.
```
tommydeboer, line 35:
```suggestion
You need to start several backend services in order to be able to develop in the Armadillo.
```

PR #178, opened 2022-07-20
feat: add user interface to armadillo

mswertz:


mswertz:
propose to merge so we can more easily contribute bits and pieces.

marikaris (approved):

PR #194, opened 2022-07-27
Add pre-commit that auto-formats code on commit

tommydeboer:
Let's try using `pre-commit` to manage our commit hooks. This one triggers a `mvn fmt:format` on commit.

1. Install `pre-commit` with `pip` or `brew`  (see https://pre-commit.com/index.html#1-install-pre-commit)
2. In the repo, run `pre-commit install`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+  -   repo: https://github.com/ejba/pre-commit-maven
+      rev: v0.3.3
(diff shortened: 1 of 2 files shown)


mswertz (changes requested):
Bit laborious to do first time (unbelievable how much stuff brew installed) but then works fine. I would recommend you add description to the README to explain this step to new developers.

mswertz (approved):

PR #197, opened 2022-07-28
chore: migrate to java 17

tommydeboer:
~TODO: figure out which image to get from https://console.cloud.google.com/gcr/images/distroless/GLOBAL/~

See: https://github.com/GoogleContainerTools/jib/blob/master/docs/default_base_image.md#migration-from-pre-30
See: https://stackoverflow.com/questions/67130671/why-jib-dropped-support-for-distroless-base-image

We will follow jib's recommendations and let jib choose the base image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,10 @@
 pipeline {
     agent {
         kubernetes {
-            inheritFrom 'molgenis-jdk11'
+            // the shared pod template defined in the Jenkins server config
+            inheritFrom 'shared'
+            // maven jdk17 pod template defined in this repository
+            yaml libraryResource("pod-templates/maven-jdk17.yaml")
         }
     }
     environment {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,7 +3,7 @@ pipeline {
         kubernetes {
             // the shared pod template defined in the Jenkins server config
             inheritFrom 'shared'
-            // maven jdk17 pod template defined in this repository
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -1,7 +1,10 @@
 pipeline {
     agent {
         kubernetes {
-            inheritFrom 'molgenis-jdk11'
+            // the shared pod template defined in the Jenkins server config
+            inheritFrom 'shared'
+            // maven jdk17 pod template defined in this repository
+            yaml libraryResource("pod-templates/maven-jdk17.yaml")
```
mswertz, line 7: Is not in this repository?

armadillo/pom.xml
```diff
@@ -189,11 +191,6 @@
           </execution>
         </executions>
         <configuration>
-          <from>
-            <!-- Current java 11 image.
-            See https://console.cloud.google.com/gcr/images/distroless/GLOBAL/java?gcrImageListsize=30 -->
-            <image>gcr.io/distroless/java@sha256:7fc091e8686df11f7bf0b7f67fd7da9862b2b9a3e49978d1184f0ff62cb673cc</image>
-          </from>
           <to>
             <image>${dockerfile.repository}:${dockerfile.tag}</image>
```
mswertz, line 195: this doesn't make sense to me: here I would expect a normal docker image?

mswertz (changes requested):

mswertz (approved):

PR #198, opened 2022-07-28
feat: added API for local access permission granting

mswertz:
Todo:

- [x] ability to grant/revoke su
- [x] test for JwtDecoder
- [x] ability to store user details like firstname, lastname, institution (to: can this be automatic based on oauth info?)

To test: 
- set client secret in application.yml
- then you can sign in using fusion auth
- then you can get a token via http://localhost:8080/my/token
- you can also use swagger-ui
- you need to add your account to 'administrators' project to be able to use UI
- therefore you can also sign in via basic auth (so you can add yourself as administrator, and then 
- POST /access {email: your email, project: admininistrators}
- after that you could use JWT auth as well

N.B. we need better docs, propose to add a 'docs' folder and then integrate analogous to emx2 docs.


mswertz:
[local grant/revoke access in Armadillo instead of depending on FusionAuth roles](https://trello.com/c/VKymyCtp/924-local-grant-revoke-access-in-armadillo-instead-of-depending-on-fusionauth-roles)

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettingsService.java
```diff
@@ -0,0 +1,115 @@
...
+
+  @PreAuthorize("hasRole('ROLE_SU')")
+  public void userDelete(String email) {
+    Objects.requireNonNull(email);
+    reloadIfNeeded();
+    settings.getUsers().remove(email);
+    save();
+  }
+
+  public synchronized void save() {
```
tommydeboer, line 92: can be private

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettingsService.java
```diff
@@ -0,0 +1,115 @@
...
+    String json = new Gson().toJson(settings);
+    InputStream inputStream = new ByteArrayInputStream(json.getBytes());
+    armadilloStorageService.saveSystemFile(inputStream, SETTINGS_FILE, MediaType.APPLICATION_JSON);
+    forceReload = true;
+  }
+
+  public void reloadIfNeeded() {
```
tommydeboer, line 99: can be private

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettingsService.java
```diff
@@ -0,0 +1,115 @@
...
+
+@Service
+public class ArmadilloSettingsService {
+
+  public static final String SETTINGS_FILE = "users.json";
+  private ArmadilloSettings settings;
+  @Autowired private ArmadilloStorageService armadilloStorageService;
+  private boolean forceReload = true;
+
+  @PreAuthorize("hasRole('ROLE_SU')")
```
tommydeboer, line 22: We can also put this on the class level to protect everything at once

armadillo/src/main/java/org/molgenis/armadillo/settings/UserDetails.java
```diff
@@ -0,0 +1,55 @@
+package org.molgenis.armadillo.settings;
+
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
+
+// for generation to JSON
+public class UserDetails {
```
tommydeboer, line 8: We use `@AutoValue` for objects that need to be serialised to JSON. See for examples `Workspace` or `RPackage`

armadillo/src/test/java/org/molgenis/armadillo/TestSecurityConfig.java
```diff
@@ -13,4 +21,22 @@ public class TestSecurityConfig extends WebSecurityConfigurerAdapter {
...
+  ArmadilloSettingsService accessStorageService() {
+    return new ArmadilloSettingsService();
+  }
+
+  @Bean
+  // used in SettingsController test
+  public UserDetailsService userDetailsService() {
+    GrantedAuthority authority = new SimpleGrantedAuthority("ROLE_SU");
+    User userDetails =
+        new User("bofke", "bofke", Arrays.asList(authority)) {
```
tommydeboer, line 35: 👍

pom.xml
```diff
@@ -1,266 +1,273 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
tommydeboer, line 1: Fix formatting: I can't see if anything has changed

armadillo/src/main/java/org/molgenis/armadillo/security/AuthConfig.java
```diff
@@ -2,61 +2,173 @@
...
-        .and()
-        .httpBasic()
-        .realmName("Armadillo")
-        .and()
-        .oauth2ResourceServer(
-            oauth2 ->
-                oauth2.jwt(jwt -> jwt.jwtAuthenticationConverter(grantedAuthoritiesExtractor())));
+  @Configuration
+  @EnableWebSecurity
+  @Profile({"armadillo", "development"})
```
tommydeboer, line 45: Shouldn't this say `production` instead of `armadillo`? In a few other places as well

armadillo/src/main/java/org/molgenis/armadillo/security/AuthConfig.java
```diff
@@ -2,61 +2,173 @@
...
-    JwtAuthenticationConverter jwtAuthenticationConverter = new JwtAuthenticationConverter();
-    jwtAuthenticationConverter.setJwtGrantedAuthoritiesConverter(new JwtRolesExtractor());
-    return jwtAuthenticationConverter;
+  @Configuration
+  @EnableWebSecurity
+  @Order(2)
+  @ConditionalOnProperty(
+      value = "spring.security.oauth2.client.registration.molgenis.client-id",
+      matchIfMissing = true,
+      havingValue = "value_that_never_appears")
```
tommydeboer, line 101: `havingValue` can be left out I believe

tommydeboer (changes requested):

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettings.java
```diff
@@ -0,0 +1,14 @@
+package org.molgenis.armadillo.settings;
+
+import com.google.auto.value.AutoValue;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+@AutoValue
+public abstract class ArmadilloSettings {
+  public abstract Map<String, UserDetails> getUsers();
```
tommydeboer, line 9: We need a concurrent map here because multiple threads can read/write to the settings.

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettingsService.java
```diff
@@ -0,0 +1,114 @@
...
+    armadilloStorageService.saveSystemFile(inputStream, SETTINGS_FILE, MediaType.APPLICATION_JSON);
+    forceReload = true;
+  }
+
+  private void reloadIfNeeded() {
+    if (forceReload) {
+      InputStream inputStream = armadilloStorageService.loadSystemFile(SETTINGS_FILE);
+
+      ArmadilloSettings temp =
+          new Gson().fromJson(new InputStreamReader(inputStream), ArmadilloSettings.class);
```
tommydeboer, line 103: `Gson` object can be initialised once in constructor

tommydeboer (changes requested):

tommydeboer:
**Issue:** The test script (`docker/test/test-local.R`) can't authenticate anymore because it uses basic auth. See https://github.com/molgenis/molgenis-r-datashield/blob/master/R/ArmadilloDriver.R#L23-L101

If we disable basic auth we need to update the client's driver

tommydeboer:
**Possible issue:**

From the logging: 

```
09:23:11.964 [main|] INFO  o.s.s.web.DefaultSecurityFilterChain - Will not secure And [Not [Ant [pattern='/']], Not [Ant [pattern='/oauth2/**']], Not [Ant [pattern='/login/**']], Not [Ant [pattern='/login']]]
09:23:11.998 [main|] INFO  o.s.s.web.DefaultSecurityFilterChain - Will not secure any request
```

tommydeboer:
**Issue:**
With OIDC enabled, I can not use the form login:

image

tommydeboer:
**Issue:**
After logging in with login form, and calling the `GET /users` endpoint, I get:

```
{json
  "timestamp": "2022-08-04T07:43:11.298+00:00",
  "status": 500,
  "error": "Internal Server Error",
  "message": "Cannot invoke \"org.molgenis.armadillo.settings.ArmadilloSettings.getUsers()\" because \"this.settings\" is null",
  "path": "/users"
}
```

MinIO is empty so something's wrong with the persistence. Maybe same issue you had yesterday with the users.json not updating?

armadillo/src/main/java/org/molgenis/armadillo/settings/ArmadilloSettings.java
mswertz: Good point. I added some synchronized on the file writing bits, don't know if that helps in reality.
In practice this is a single user app so we shouldn't worry so much, but will make this fix of course :-)

mswertz:
> If we disable basic auth we need to update the client's driver

Okay, will try to switch back to basic auth then.

mswertz:
> **Possible issue:**
> 
> From the logging:
> 
> ```
> 09:23:11.964 [main|] INFO  o.s.s.web.DefaultSecurityFilterChain - Will not secure And [Not [Ant [pattern='/']], Not [Ant [pattern='/oauth2/**']], Not [Ant [pattern='/login/**']], Not [Ant [pattern='/login']]]
> 09:23:11.998 [main|] INFO  o.s.s.web.DefaultSecurityFilterChain - Will not secure any request
> ```

I think these are the 'monitoring' endpoints that are not secured

tommydeboer:
**Issue:** 
I started Armadillo without OIDC configured. On the start screen I went to _Login using institute account (oauth2)._ but this sent me directly to the Swagger UI, instead of showing me a login screen. 

Maybe I was still logged in? So I went back and clicked _Logout_. Nothing seems to happen. If I click _Login using institute account (oauth2)._ again I am sent directly to Swagger UI again.

tommydeboer:
**Issue:**
Local permissions can still be overridden by roles in the JWT.

mswertz:
I think we are nearly done? We only need to decide on the uppercase project name thing, see above.

tommydeboer (approved):
Everything's working nicely :) 

Concerning the project/object validation rules: I propose to make it part of the Storage API PR. We can decide on Monday which API has which responsibility

PR #201, opened 2022-08-05
Storage API

tommydeboer:
TODO:

- [x] add better swagger descriptions
- [x] implement minio delete project
- [x] what to do with storage methods in `DataController`?
- [x] auditing
- [x] unit tests
- [x] use global openapi definition 
- [x] rename /metadata to /admin
- [x] create metadata + folder when creating project|
- [x] global project/object validation rules + exceptions
- [x] secure all service endpoints and introduce runAsSystem


armadillo/http/test-storage.http
```diff
@@ -0,0 +1,51 @@
...
+POST {{host}}/storage/projects/lifecycle/objects
+Content-Type: multipart/form-data; boundary=boundary
+
+--boundary
+Content-Disposition: form-data; name="file"; filename="core/dummy.parquet"
+
+< ./dummy.parquet
+
+### Copy object
+POST {{host}}/storage/projects/lifecycle/objects/core%2Fnonrep.parquet/copy
```
mswertz, line 33: huh, do we really need copy and move? is that not a bit over engineering?

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloServiceApplication.java
```diff
@@ -3,12 +3,28 @@
...
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.retry.annotation.EnableRetry;
+import org.springframework.web.servlet.config.annotation.PathMatchConfigurer;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+import org.springframework.web.util.UrlPathHelper;
 
 @SpringBootApplication(scanBasePackages = "org.molgenis")
 @EnableRetry
-public class ArmadilloServiceApplication {
+public class ArmadilloServiceApplication implements WebMvcConfigurer {
```
mswertz, line 12: Don't know if it matters, but you could also use https://github.com/molgenis/molgenis-service-armadillo/blob/master/armadillo/src/main/java/org/molgenis/armadillo/interceptor/WebConfig.java which is also a WebMvcConfigurer?

Generally, this is why I don't like Spring. To much indirection.

armadillo/src/main/java/org/molgenis/armadillo/controller/ControllerExceptionHandler.java
```diff
@@ -0,0 +1,51 @@
...
+import org.molgenis.armadillo.exceptions.DuplicateProjectException;
+import org.molgenis.armadillo.exceptions.FileProcessingException;
+import org.molgenis.armadillo.exceptions.StorageException;
+import org.molgenis.armadillo.exceptions.UnknownObjectException;
+import org.molgenis.armadillo.exceptions.UnknownProjectException;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+
+@ControllerAdvice
```
mswertz, line 18: Ah, that is interesting, a form of centralized error handling?

armadillo/src/main/java/org/molgenis/armadillo/controller/ProjectRequestBody.java
```diff
@@ -0,0 +1,5 @@
+package org.molgenis.armadillo.controller;
+
+import javax.validation.constraints.NotNull;
+
+public record ProjectRequestBody(@NotNull String name) {}
```
mswertz, line 5: overlaps with permission API.

armadillo/src/main/java/org/molgenis/armadillo/controller/ObjectRequestBody.java
```diff
@@ -0,0 +1,5 @@
+package org.molgenis.armadillo.controller;
+
+import javax.validation.constraints.NotNull;
+
+public record ObjectRequestBody(@NotNull String name) {}
```
mswertz, line 5: Why use here a record why elsewhere we use @AutoValue?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,158 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.multipart.MultipartFile;
+
+@OpenAPIDefinition(
```
mswertz, line 41: should be removed to central OpenApiConfig, see other PR.

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,158 @@
...
+      @SecurityRequirement(name = "http"),
+      @SecurityRequirement(name = "jwt")
+    })
+@SecurityScheme(name = "JSESSIONID", in = COOKIE, type = APIKEY)
+@SecurityScheme(name = "http", in = HEADER, type = HTTP, scheme = "basic")
+@SecurityScheme(name = "jwt", in = HEADER, type = APIKEY)
+@RestController
+@Validated
+@PreAuthorize("hasRole('ROLE_SU')")
+@RequestMapping("storage")
```
mswertz, line 54: careful: if we change mappings used in DSI we must also create new release of ArmadilloDSI package (and all packages that depend on it)?

Do we worry about that?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,158 @@
...
+  @GetMapping("/projects")
+  @ResponseStatus(OK)
+  public List<String> listProjects() {
+    return storage.listProjects();
+  }
+
+  @PostMapping(
+      value = "/projects",
+      consumes = {APPLICATION_JSON_VALUE})
+  @ResponseStatus(NO_CONTENT)
```
mswertz, line 72: duplicate with other PR as you have seen. I feel we should merge asap.

mswertz (commented):
looks already good. This .http language, does that also have a mechanism to test that outcomes are equal to some result?

Further comments inline

mswertz:
[Extend data api for CRUD on projects, folders, files](https://trello.com/c/01tp1oqV/929-extend-data-api-for-crud-on-projects-folders-files)

armadillo/http/test-storage.http
tommydeboer: They were present in the proposal I sent you. The client does provide copy and move, and my thinking was: better to let the backend manage this than spreading the operation over multiple requests and making the client download and upload the file again.

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloServiceApplication.java
tommydeboer: Good point, forgot to look for another implementation

armadillo/src/main/java/org/molgenis/armadillo/controller/ControllerExceptionHandler.java
tommydeboer: Yes this is the "preferred" way of exception handling in controllers because it centralises the exceptions and gives you a lot of control over the responses.

tommydeboer: Btw, the best way to return errors to the client would be to wrap them together with some other values, for example:

```json
{
  "error": 
  {
    "code": "404",
    "date": "12-08-2022'09:21"
    "message": "error message"
    "stacktrace": "stacktrace when in development mode"
  }
}
```

This controller advice give us a nice setup to add something like that easily in the future.

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: Yes I was keeping it for after the merge to prevent conflicts

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: There are a few methods of which I first thought that they should be moved to the storage API. (The ones that are marked with TODO in `DataController`). However, these endpoints behave differently: they list things from a "datashield" point of view. For example, listing all tables you can access as a user across all projects you can access. The Storage API works on a "single-project" level.

We need to keep both APIs so I will remove the TODOs. I want to refactor the `ArmadilloStorageService` a bit (maybe split it up) to make this distinction more clear.

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: agree, let's focus on the other PR first

armadillo/src/main/java/org/molgenis/armadillo/controller/ObjectRequestBody.java
tommydeboer: See discussion on Slack

tommydeboer:
> This .http language, does that also have a mechanism to test that outcomes are equal to some result?

Yes, see [the HTTP client docs](https://www.jetbrains.com/help/idea/http-response-handling-examples.html#checking-response-headers-body)

There is also an [open source CLI runner](https://github.com/restcli/restcli) that can run these tests independent of the IDE so they can be used in CI builds. I would like to explore this as a candidate for the system tests.

README.md
```diff
@@ -62,7 +62,15 @@ If you want to use MinIO as storage (including the test data), do the following:
 4. Add a bucket `shared-lifecyle`
 5. Copy the folders in `data/shared-lifecycle` in this repository to the bucket
 6. In `application.yml`, uncomment the `minio` section.
-7. Now Armadillo will automatically connect to MinIO at startup. 
+7. Now Armadillo will automatically connect to MinIO at startup.
```
mswertz, line 65: might want to describe the 'local' version too?

mswertz (commented):
code review looks good, next step is that I will run and test.

mswertz:
We should address the sonarcloud issues.

README.md
tommydeboer: See the beginning of this chapter 😉

armadillo/src/main/java/org/molgenis/armadillo/controller/ControllerExceptionHandler.java
tommydeboer: And just remembered that with the `@ResponseStatus` annotations on exceptions above is no longer needed 😄

armadillo/src/main/java/org/molgenis/armadillo/audit/AuditEventPublisher.java
```diff
@@ -95,46 +104,72 @@ public void audit(
...
     try {
       var result = c.get();
       audit(principal, type, data);
       return result;
     } catch (Exception failure) {
-      Map<String, Object> errorData = new HashMap<>(data);
-      errorData.put(MESSAGE, failure.getMessage());
-      errorData.put(TYPE, failure.getClass().getName());
-      audit(principal, type + "_FAILURE", errorData);
+      auditFailure(principal, type, data, failure);
```
mswertz, line 134: nice this auditFailure wrapper.

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
mswertz: good

armadillo/src/main/java/org/molgenis/armadillo/exceptions/InvalidProjectNameException.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(BAD_REQUEST)
+public class InvalidProjectNameException extends RuntimeException {
+
+  public InvalidProjectNameException(String projectName) {
+    super(
+        format(
+            "Project name '%s' is invalid. Names should adhere to the rules described at "
```
mswertz, line 14: I don't mind that we do it, but I just want to note that me that creating classes for each variant of exception look a like a bit 'code heavy' in my eyes. Rather I would prefer a few high level exceptions that are parameterized with the right message (and if you really want to standardize those message to use a central repository of messages). But I do understand this is a 'perfect pattern' that works well. (you are probably irritated by the lack of diversity in exceptions in emx2)

armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
```diff
@@ -133,4 +200,31 @@ public InputStream loadSystemFile(String name) {
       return InputStream.nullInputStream();
     }
   }
+
+  private void throwIfDuplicate(String project, String object) {
```
mswertz, line 204: why do you use this intermediate function instead of throwing the exception directly at each spot?

mswertz (commented):
code review looks good, will now move to actually running

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,292 @@
...
+
+  private final ArmadilloStorageService storage;
+  private final AuditEventPublisher auditor;
+
+  public StorageController(ArmadilloStorageService storage, AuditEventPublisher auditor) {
+    this.storage = storage;
+    this.auditor = auditor;
+  }
+
+  @Operation(summary = "List all projects")
```
mswertz, line 72: I would propose to remove this one in favour of the /admin/projects that has more information. What is use case to duplicate?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,292 @@
...
+      })
+  @GetMapping(
+      value = "/projects",
+      produces = {APPLICATION_JSON_VALUE})
+  @ResponseStatus(OK)
+  public List<String> listProjects(Principal principal) {
+    return auditor.audit(storage::listProjects, principal, LIST_PROJECTS, Map.of());
+  }
+
+  @Operation(summary = "Project exists?")
```
mswertz, line 93: what is use case of this 'head' on projects? Seems to be unneeded?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,292 @@
...
+    return auditor.audit(
+        () -> storage.listObjects(project), principal, LIST_OBJECTS, Map.of(PROJECT, project));
+  }
+
+  @Operation(summary = "Upload an object to a project")
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "204", description = "Object uploaded successfully"),
+        @ApiResponse(responseCode = "404", description = "Unknown project"),
+        @ApiResponse(responseCode = "409", description = "Object already exists"),
```
mswertz, line 139: we could consider to convert from 'post' to 'put'. Then you would overwrite existing. Might be more convenient? In the admin API we implemented this upsert pattern (which made much sense in light that we there really update a json blob we had to have update functionality)

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
mswertz: Btw, I get impression your quality of annotation is much better then what I did before on AdminController. I can imagine you would want to also put some TLC in the other controllers to make them just as well documented?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -0,0 +1,292 @@
...
+        @ApiResponse(responseCode = "204", description = "Object copied successfully"),
+        @ApiResponse(responseCode = "404", description = "Unknown project or object"),
+        @ApiResponse(responseCode = "409", description = "Object already exists"),
+        @ApiResponse(responseCode = "401", description = "Unauthorized")
+      })
+  @PostMapping(
+      value = "/projects/{project}/objects/{object}/copy",
+      consumes = {APPLICATION_JSON_VALUE})
+  @ResponseStatus(NO_CONTENT)
+  public void copyObject(
```
mswertz, line 178: I must confess this operation confused me. I do understand we then need to post a new name for the rename but it feels a bit clunky (aggrevated by swagger, where I simply expected an 'oldName' and 'newName' parameter instead of the third parameter being a json object.

Moreover, I can imagine users would like to copy or move to another project? So we might want to expand requestbody to include project as optional?

mswertz (commented):
also after testing code looks good. A few comments I would like your view on before approving.

armadillo/src/main/java/org/molgenis/armadillo/exceptions/InvalidProjectNameException.java
tommydeboer: I am very very in favour of this method and I actually think this is one of the things MOLGENIS got right 😉 . Here's why:

- It improves readability because the code is not littered with messages
- It simplifies writing exception handling because there is a list of exceptions to choose from
- It ensures consistency of error messages because people don't need to think of one every time they throw an exception
- The exceptions can actually be acted on: for example, we return a `409` on `DuplicateProjectException`, and a `404` on `UnknownProjectException`. This benefit is lost with general exceptions.
- It doesn't take more than a minute to create a new exception
- I would actually consider this method less 'code heavy', because you don't pollute the actual code with long and often duplicated strings

Some of these problems can also be solved by a central repository but then people still are free to use whatever message they want (which in my eyes is a bad thing). If you want to support i18n in the future, passing strings to exceptions will be a pain to refactor (as we have experienced with MOLGENIS 😬).

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: Hmm, I haven't considered it, but it sounds a bit dangerous to me. You could delete/overwrite data you didn't intend to without you knowing about it. I think doing one extra request is worth the added security.

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: I agree, this one isn't the easies to use (in Swagger) :/ It's because of the mix of path values and request body. I don't really see a better way, do you?

Good point about the copying to another project. Do you want to do it now? We can also wait until it's requested because adding an optional request body paramater will be backwards compatible.

tommydeboer: Hmmm, how about

`/storage/projects/{project}/objects/{object}/copyTo/{object}`

this wouldn't work for copying accross projects though, but would simplify usage (in Swagger) greatly

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: Ah yes, good point. It didn't consider removing it yet. The endpoint does make a lot of "REST" sense, but it might confuse people about which API to use. Removing 👍

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
tommydeboer: More out of completeness I think, I thought it might come in handy in the client. But since we raise a `404` when the project doesn't exist, it's not really needed

tommydeboer: Yes I had fun exploring the annotations a bit 😄  I agree it would be nice to have such extensive documentation on other APIs as well, I'll take a look

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
mswertz: ok

mswertz: ok

armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
mswertz: Ah yes, I should have looked better.

mswertz (approved):

PR #230, opened 2022-09-21
Migrations scripts for Armadillo 2.x -> 3.0

tommydeboer:


scripts/migrate-auth.py
```diff
@@ -0,0 +1,192 @@
...
+    # 1. Set up clients
+    armadillo_client = create_armadillo_client()
+    fusion_client = create_fusion_client()
+
+    # 2. Get production apps that have users
+    users = get_active_registered_users(fusion_client)
+    apps = get_production_applications_with_users(fusion_client, users)
+
+    # 3. Let user pick which app to migrate to this Armadillo server
+    app_to_migrate = choose_application(apps)
```
mswertz, line 63: Did you test this with a 'data manager' user? I.e. what permissions must a user have to do this? Data managers have 'SU' in only one or more applications, would they also be able to do this?

mswertz: Oh, and what happens when user doesn't have any applications to choose from?

mswertz (approved):
okay, lovely code. Just two small questions you might one to check before merge.

scripts/migrate-auth.py
tommydeboer: As discussed, we will use API key for now and if it happens that a project's SU wants to do the migration theirselves, then we will add that option later.

tommydeboer:
[Migration script for Fusion auth [blocks on cluster]](https://trello.com/c/nG3erycy/945-migration-script-for-fusion-auth-blocks-on-cluster)

PR #233, opened 2022-09-26
Fix: App won't start with non-existing local filestorage dir

tommydeboer:


tommydeboer:
[Fix: Armadillo still checks local file storage path when MinIO is configured](https://trello.com/c/0B5Q3wA5/1039-fix-armadillo-still-checks-local-file-storage-path-when-minio-is-configured)

PR #311, opened 2023-01-17
Installation script

DickPostma:


scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+          ;;
+        *)
+        break;
+        esac
+        
+    done
+    if [ ! "$ARMADILLO_CLEANUP" ]; then
+      if [ ! "$ARMADILLO_DOMAIN" ] || [ ! "$ARMADILLO_ADMIN_PW" ]; then
+        echo "Arguments --domain --admin-password must be provided"
+        echo "U need an host or domain to use Armadillo: Example: cohort.armadillo.organisation.com"
```
tommydeboer, line 74:
```suggestion
        echo "You need a host or domain to use Armadillo. Example: cohort.armadillo.organisation.com"
```

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+        echo "Also for security reasons you must provide an secure admin password"
```
tommydeboer, line 75:
```suggestion
        echo "Also for security reasons you must provide a secure admin password"
```

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    echo
+    echo 'Params:'
+    echo
+    echo '    --version armadillo_version   Specify witch version to install'
+    echo '    --admin-user user             Specify the Basic-Auth admin user'
+    echo '    --admin-password pass         password for the admin user'
+    echo '    --datadir /storage/dir        If defined this would be the Location to store the data otherwise it would be /usr/share/armadillo/data'
+    echo '    --domain                      URL domain witch is used for accessing armadillo'
+    echo ''
+    echo '    --oidc                          For central authentication you can enable oidc'
```
tommydeboer, line 269: OIDC is required so I think this option is not necessary

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    useradd -rs /bin/false "$ARMADILLO_SYS_USER"
+    chgrp -R "$ARMADILLO_SYS_USER" "$ARMADILLO_PATH"
+    chgrp -R "$ARMADILLO_SYS_USER" "$ARMADILLO_CFG_PATH"
+    chgrp -R "$ARMADILLO_SYS_USER" "$ARMADILLO_LOG_PATH"
+    chgrp -R "$ARMADILLO_SYS_USER" "$ARMADILLO_DATADIR"
+
+    chmod g+rw "$ARMADILLO_LOG_PATH"
+    chmod g+rw "$ARMADILLO_DATADIR"
+    usermod -aG docker "$ARMADILLO_SYS_USER"
+    echo "Environment is beeing setup correctly"
```
tommydeboer, line 106:
```suggestion
    echo "Environment is being set up correctly"     
```

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+
+EOF
+
+systemctl daemon-reload
+echo "Armadillo Installed under systemd"
+}
+
+setup_armadillo_config() {
+  if [ "$ARMADILLO_OIDC_ENABLED" ]; then
+    wget -q -O $ARMADILLO_CFG_PATH/application.yml https://raw.githubusercontent.com/molgenis/molgenis-service-armadillo/scripts/conf/application-oidc.yml
```
tommydeboer, line 140: Do you think we should pin these URLs to a specific version somehow? Now the script will always get the latest version from the dev/master branch. We might make changes there before a new version has been released, potentially breaking older installations. Another strategy is to always make this installation script backwards compatible. Maybe not something we need to fix ASAP, but something to keep in mind for the future.

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    if [ $? -ne 0 ]; then
+        echo "I require $1 but it's not installed. Abort."
+        exit 1
+    fi
+}
+
+
+cleanup(){
+  if [ "$ARMADILLO_CLEANUP" ]; then
+    echo "--cleanup variable defined! Are you sure. Potential data loss? Type y/N"
```
tommydeboer, line 224:
```suggestion
    echo "--cleanup variable defined! Are you sure? Potential data loss! Type y/N"
```

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+
+startup_armadillo() {
+  systemctl enable armadillo
+  systemctl start armadillo
+  echo "Armadillo started"
+
+}
+
+#Parameters passed in help
+parameters_help() {
```
tommydeboer, line 254: The `--cleanup` argument is missing in the help

scripts/conf/application-oidc.yml
```diff
@@ -0,0 +1,94 @@
+armadillo:
+  # set this to 'true' if you want to accept permissions from oidc provider
+  oidc-permission-enabled: false
+  # set this true if you want Armadillo to create/edit/delete profile docker images
+  docker-management-enabled: true
+  # uncomment this to configure a default admin user
+  # oidc-admin-user: user@yourdomain.org
+          # foo: bar
```
tommydeboer, line 8:
```suggestion
```

scripts/conf/application-oidc.yml
```diff
@@ -0,0 +1,94 @@
...
+
+storage:
+  root-dir: @DATADIR@
+
+# Uncomment the following block to enable MinIO storage (overrules local storage)
+# minio:
+#   url: http://minio
+#   port: 9000
+#   access-key: ASSDFXrJoymrOZ8rp
+#   secret-key: ASDFADkC87qmM99O
```
tommydeboer, line 94: Let's put some placeholders instead of keys here

scripts/conf/application.yml
```diff
@@ -0,0 +1,79 @@
+armadillo:
```
tommydeboer, line 1: I don't think this file is needed since OIDC is required

tommydeboer (changes requested):
Looking good, but I have a few questions and have fixed some typos.

Shall we also put these files under `script/install` instead of under `script` to keep it nice and organised?

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    echo 'Usage: ./armadillo.sh PARAMS 
+       example ./armadillo --admin-user admin --admin-password welcome01 --domain armadillo.cohort.study.com'
+    echo
+    echo 'Install Script for Armadillo Service'
+    echo
+    echo 'Params:'
+    echo
+    echo '    --version armadillo_version   Specify witch version to install'
+    echo '    --admin-user user             Specify the Basic-Auth admin user'
+    echo '    --admin-password pass         password for the admin user'
```
marikaris, line 265: Password with a capital P

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    echo '    --datadir /storage/dir        If defined this would be the Location to store the data otherwise it would be /usr/share/armadillo/data'
```
marikaris, line 266: Add comma:
 If defined this would be the Location to store the data, otherwise it would be /usr/share/armadillo/data'

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    echo '    --domain                      URL domain witch is used for accessing armadillo'
```
marikaris, line 267: witch = which

scripts/armadillo-setup.sh
```diff
@@ -0,0 +1,298 @@
...
+    echo ''
+    echo '    --oidc                          For central authentication you can enable oidc'
+    echo '      --oidc_url                    URL where the oidc server is listening on'
+    echo '      --oidc_clientid               client id of the oidc config'
+    echo '      --oidc_clientsecret           secret of the client'
+
```
marikaris, line 273: On line 271 Client with capital C
On line 272 Secret with capital S

scripts/conf/application.yml
```diff
@@ -0,0 +1,79 @@
...
+  # set this to 'true' if you want to accept permissions from oidc provider
+  oidc-permission-enabled: false
+  # set this true if you want Armadillo to create/edit/delete profile docker images
+  docker-management-enabled: true
+  # uncomment this to configure a default admin user
+  # oidc-admin-user: user@yourdomain.org
+          # foo: bar
```
marikaris, line 8: remove?

scripts/armadillo-setup.sh
mswertz: Then we should add both script and yaml to the releases. Then you can get it there. Or use tags and get it from github repo of a particular version. Or more easy, use simply the tags.

mswertz: BTW: I plan to change Armadillo repo to behave similar to emx2, which means any merge into the master is also a release. In that case master is always latest release and then this is generally not a problem.

mswertz (commented):
Nice work. Much more advanced than I expected (I expected simply a 'download' and 'run', not setting up the full monty with ngingx and as a service).

I miss an addition to the README.md how to use this script. And we should say this aims at installing Armadillo as a service on a Linux machine. In that light, current script seems a bit heavy for users that just want to play with it. I would love also a lighter version of the script that just downloads and runs.

Ideally README.md would read:

To run Armadillo you need recent Java and Docker. You can install and run Armadillo in your current working directory or as a system service (currently Linux and Apple only).

To install and run download 'install.sh' from a release at https://github.com/molgenis/molgenis-service-armadillo/releases

To run locally on Linux or Mac:

```sh install.sh```

To install and run as a service on Linux server

```sh install.sh --service```

To see what parameters are available

````sh install.sh --help``

File: scripts/armadillo-setup.sh [hunk 16]
```diff
@@ -0,0 +1,298 @@
...
+
+handle_args() {
+    while :
+    do
+        case "$1" in 
+        --version)
+          ARMADILLO_VERSION=$2
+          shift 2
+          ;;
```
tommydeboer, line 21: I think it's a good idea to add `--admin-email` here as well. This will allow the admin to immediately login with OIDC instead of first having to login with basic auth, then add the email, and reauthenticate.

tommydeboer: See `oidc-admin-user` in the config

scripts/armadillo-setup.sh
tommydeboer: Have to correct myself here: OIDC is not required, however configuring a resource server for JWT authentication **is**   mandatory.

scripts/armadillo-setup.sh
DickPostma: At my opinion is that configuration is an one time only process at initial setup. After installation the config files shouldn't be updated anymore (not automatically). So when you upgrade armadillo and add config features it should not be mandatory. (Or it has to be a major release, where you instruct users to update their config)

scripts/armadillo-setup.sh
DickPostma: The idea was that this shoud be an hidden feature.. (Describe it in the documentation as last resort of uninstall part, or use it as an DevOps process)

christianhilbrands (commented):
RfC

README.md
```diff
@@ -43,7 +43,29 @@ federates to authentication systems of connected institutions, ideally using a f
...
 # Getting started
 
-To spin up your own server on a laptop, you can run `java -jar armadillo-3.x.x.jar`
+## Prerequisites
+
+Armadillo 3 can be installed on any flavor of linux OS or modern Unix-based Mac OS. 
+To run armadillo3 you need the following dependencies.
+
+* Java 17 JRE or JDK
+* Docker (for profiles)
```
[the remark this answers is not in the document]
DickPostma: Bash is standard installed in Ubuntu

scripts/install/README.md
```diff
@@ -0,0 +1,72 @@
...
+> | 0-20.000      | 8              | 100               | 4         |
+> | 20.000-70.000 | 16             | 100               | 4         |
+> | 70.000 >      | 32             | 150               | 8         |
+
+In case of using dsOmics this setup can be rather bigger. Please contact molgenis-operations@umcg.nl for the latest specifications.
+
+## Software requirements
+
+* Java 17 JRE or JDK
+* Docker
```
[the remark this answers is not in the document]
DickPostma: Bash is standard installed in Ubuntu

PR #357, opened 2023-02-07
Fix/#314 highlight edit mode

marikaris:


ui/src/views/Projects.vue
```diff
@@ -190,6 +191,9 @@ export default defineComponent({
     };
   },
   computed: {
+    editHighlight(): "info" | "" {
```
tommydeboer, line 194: At the moment we only do an "info" highlight so you can move that to the `InlineRowEdit` component. Then you can just pass the `isEditingProject` value which removes the need for this extra method. Same goes for the other components.

tommydeboer (changes requested):

tommydeboer (approved):

PR #360, opened 2023-02-07
feat(profiles): ask for confirmation when renaming profile

marikaris:


ui/src/views/Profiles.vue
```diff
@@ -16,6 +16,15 @@
           @proceed="proceedDelete"
           @cancel="clearRecordToDelete"
         ></ConfirmationDialog>
+        <ConfirmationDialog
+          v-if="isRenamingProfile"
+          :record="profileToEdit"
+          action="rename"
+          recordType="profile"
+          extraInfo="Renaming profile will cause profile with the old name to be deleted. All its running processes will be stopped."
```
tommydeboer, line 24:
```suggestion
          extraInfo="Renaming a profile will cause the old profile to be deleted. All its running processes will be stopped."
```

tommydeboer (approved):

PR #400, opened 2023-03-20
fix: Update readme and cron updates weekly

DickPostma:


scripts/install/README.md
```diff
@@ -50,7 +50,7 @@ The installation script requires some arguments:
 |oidc_secret | Given secret ID|
```
christianhilbrands, line 50: does the documentation not also have to be change to "oidc_clientsecret"?

christianhilbrands (changes requested):
RfC

scripts/install/README.md
DickPostma: Aangepast

christianhilbrands (approved):
LGTM

PR #421, opened 2023-04-18
fix: version number was not shown

mswertz:


mswertz:
sorry, merge because need to see version

PR #437, opened 2023-04-25
Refactor Rserve library usage to allow multiple R server implementations

ymarcon:
* Interfaces to define the interactions between Armadillo and a R server
* Rserve specific implementation of these interfaces : there is no more Rserve java library dependencies outside of the `org.molgenis.r.rserve` package
* Tests adapted accordingly


ymarcon:
This is a kind reminder that this PR is pending.

marikaris (approved):

PR #463, opened 2023-07-21
feat: get file info

marikaris:
Create endpoint for retrieving information from files to be included in the UI later.

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/LocalStorageService.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/LocalStorageService.java
@@ -154,6 +154,46 @@ public class LocalStorageService implements StorageService {
     }
   }
 
+  String getFileSizeInUnit(long fileSize) {
+    int sizeOfUnit = 1024;
+    String[] units = new String[] {"bytes", "KB", "MB", "GB"};
+    for (String unit : units) {
+      if (fileSize > sizeOfUnit) {
+        fileSize = fileSize / sizeOfUnit;
+      } else {
+        return fileSize + " " + unit;
+      }
+    }
+    return fileSize + " " + units[units.length - 1];
+  }
+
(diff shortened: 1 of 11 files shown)


marikaris:
[As Armadillo datamanagers I want to be able to preview resource file information](https://trello.com/c/ouykOe3U/1078-as-armadillo-datamanagers-i-want-to-be-able-to-preview-resource-file-information)

armadillo/src/main/java/org/molgenis/armadillo/storage/LocalStorageService.java
```diff
@@ -154,6 +154,43 @@ public InputStream load(String bucketName, String objectName) {
...
+  @Override
+  public Map<String, String> getInfo(String bucketName, String objectName) {
+    try {
+      Objects.requireNonNull(bucketName);
+      Objects.requireNonNull(objectName);
+
+      Path objectPath = getPathIfObjectExists(bucketName, objectName);
+      String objectPathString = objectPath.toString().toLowerCase();
+      Map<String, String> information = new HashMap<>();
+      information.put("name", objectName);
```
mswertz, line 179: You didn't want to make this a record instead of a map?

armadillo/src/main/java/org/molgenis/armadillo/storage/MinioStorageService.java
```diff
@@ -206,6 +206,11 @@ public InputStream load(String projectName, String objectName) {
     }
   }
 
+  @Override
```
mswertz, line 209: I thought you wanted to delete minio?

mswertz (approved):
I think you need to add a real test that goes through all layers?

Also, it seems we don't have sonar reporting anymore telling us about the test coverage so you can easily check if you have added sufficient tests. Something for the wish list I suppose.

PR #477, opened 2023-08-17
Fix: project explorer long names

### LINKED ISSUE #455: Long names run out of screen in project explorer view
Upload file with very long name in project explorer: 

Screenshot 2023-07-11 at 09 43 11

SietsmaRJ:
- Added function to shorten Filenames
- Added method to list-group template to check if rowIcon is table (file), if so then execute shortenFileName
- Added function to list-group-item class displayage of "key"
- Changed truncation to output `string` + `{..}` instead of `string` + `..` to better indicate truncation
- Adjusted testing for truncation change
- Added testing for shortenFileName

Closes #455

### DIFF
--- a/ui/src/components/ListGroup.vue
+++ b/ui/src/components/ListGroup.vue
@@ -14,7 +14,8 @@
           v-if="altIconCondition(key) && rowIconAlt"
           :class="`bi bi-${rowIconAlt}`"
         ></i>
-        <i v-else :class="`bi bi-${rowIcon}`"></i> {{ shortenIfFileName(key) }}
+        <i v-else :class="`bi bi-${rowIcon}`"></i>
+        {{ handleListGroupItemDisplayKey(key) }}
         <i class="bi bi-chevron-right float-end" v-if="key == selectedItem"></i>
       </a>
     </li>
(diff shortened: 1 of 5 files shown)


marikaris (changes requested):
Code looks fine. When you now however open the table preview, tables run out of the window because of the change in the truncate function. 
3 possible solutions:
1) use .. instead of {..} again
2) shorten the number of characters in the column headers
3) lessen the number of columns to be previewed
You can find an example in the lifecycle project -> core -> nonrep
Screenshot 2023-08-25 at 12 22 44

SietsmaRJ:
> Code looks fine. When you now however open the table preview, tables run out of the window because of the change in the truncate function. 3 possible solutions:
> 
> 1. use .. instead of {..} again
> 2. shorten the number of characters in the column headers
> 3. lessen the number of columns to be previewed
>    You can find an example in the lifecycle project -> core -> nonrep
>    Screenshot 2023-08-25 at 12 22 44

Couple of comments before pushing the PR feedback commit:
.1 Tried this, althrough I'm not a big fan of having a file called `somelongfi...parquet` I can't deny that this is the easiest option.
.2 I don't think that having less characters in a column header is really beneficial to the readability of the table.
.3 I also tried searching for options to make this possible, but I see that the file preview is directly from the API preview object call and then stored. I didn't really want to mess with either te API or create a function that would cut a column from each data row, which in my eyes is very resource heavy.

So I went with option 4 (or option 1.5 if 1-3 is increasing in difficulty): add a parameter to the `truncate` function, which is set to default `string = ".."`, with the shortenFileName overruling this parameter with `{..}` .

PR #478, opened 2023-08-21
Feat: datashield seed UI column

SietsmaRJ:
- Extracted Datashield Seed from Options into seperate Profiles UI column.
- Hidden datashield.seed from Options UI column.
- (To set a datashield.seed or to set a different datashield.seed, use the options column. Changing or setting a datashield seed throught the "Datashield Seed" column is disabled.)
- Added functionality to automatically asign an unused datashield seed when a new profile is created from the UI.
-

### DIFF
--- a/ui/src/views/Profiles.vue
+++ b/ui/src/views/Profiles.vue
@@ -154,6 +154,11 @@ export default defineComponent({
       profiles.value = await getProfiles()
         .then((profiles) => {
           dockerManagementEnabled.value = "container" in profiles[0];
+          for (var profile in profiles) {
+            profiles[profile]["datashieldSeed"] =
+              profiles[profile].options["datashield.seed"];
+          }
+          console.log(profiles);
           return profiles;
         })
         .catch((error: string) => {
@@ -221,6 +226,7 @@ export default defineComponent({
         port: "string",
         packageWhitelist: "array",
         functionBlacklist: "array",
+        datashieldSeed: "string",
         options: "object",
       };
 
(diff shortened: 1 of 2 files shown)


ui/src/views/Profiles.vue
```diff
@@ -213,6 +219,17 @@ export default defineComponent({
       }
       return port;
     },
+    firstFreeSeed(): number {
+      let seed = 100000000;
+      while (
+        this.profiles.find(
+          (profile) => profile.options["datashield.seed"] == seed
```
marikaris, line 226: In javascript it's `===` instead of `==` (don't ask me why 🙈)

marikaris (changes requested):
Screenshot 2023-08-25 at 12 48 25
When adding a profile, the seed is still shown under options while an empty column for the seed is shown.

ui/src/views/Profiles.vue
SietsmaRJ: There was a reason behind this, VSCode didn't like the comparison between "string" and "number"... Changing the `Ref<Profile[]>` `Profile.options["datashield.seed"]` to `number` AND `Profile.datashieldSeed` to `number` also fixes this, with the use of `===`.

SietsmaRJ: After speaking with Dick: his understanding is that at some point we want to support 64 bit integers and strings in this `options.datashield.seed` field, so that's why I would prefer to keep it `==` so that in the future when we do support strings this part doesn't clash.

StuartWheater:
Yes, DataSHIELD supporting 64bit seeded random numbers has been requested.

marikaris (approved):

PR #480, opened 2023-08-24
Feat: extract users cohorts script

SietsmaRJ:
- Created Python script to extract cohorts and users from a fusion auth url and map the users to their corresponding cohort.
- Added pandas and numpy to Pipfile

### DIFF
--- a/scripts/export-users.py
+++ b/scripts/export-users.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python3
+
 """
 User export script to aid in migration
 
--- a/scripts/export-users.py
+++ b/scripts/export-users.py
@@ -1,0 +1,144 @@
+"""
+User export script to aid in migration
+
+This script gets users and their assigned cohorts from Fusion Auth to Armadillo.
+It will group each user by cohort and will export it as TSV to a given export location, so that upon request
+this TSV can be used to set users to their appropriate cohorts.
+"""
+
+import os
+import argparse
+from pathlib import Path
+from getpass import getpass
+
+import requests
+from fusionauth.fusionauth_client import FusionAuthClient
+from fusionauth.rest_client import ClientResponse
+from requests import Session
+import pandas as pd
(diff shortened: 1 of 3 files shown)


marikaris:
[M10353 As armadillo datamanager I would like to download a list of users from the fusionauth](https://trello.com/c/9mQch7NQ/1088-m10353-as-armadillo-datamanager-i-would-like-to-download-a-list-of-users-from-the-fusionauth)

scripts/export-users.py
```diff
@@ -0,0 +1,372 @@
+#!/usr/bin/env python3
+
+"""
+User export script to aid in migration
+
+This script gets users and their assigned cohorts from Fusion Auth to Armadillo.
+It will group each user by cohort and will export it as TSV to a given export location, so that upon
+request this TSV can be used to set users to their appropriate cohorts.
```
clemens-tolboom, line 8: Isn't this "just exporting Fusion users into a TSV? Or is it a special Armadillo compatible format?

scripts/export-users.py
```diff
@@ -0,0 +1,372 @@
...
+                        roles = np.nan
+                    else:
+                        roles = self._process_roles(registration['roles'])
+                    self.cohort_dict[identifier].add_user(current_user[self.columns_of_interest])
+                    self.cohort_dict[identifier].add_user_roles(
+                        current_user['email'].values[0],
+                        roles
+                    )
+        if user_mail:
+            self.users_not_parsed.append(user_mail)
```
clemens-tolboom, line 156: Shouldn't this be placed into `for user in users...`?

clemens-tolboom (commented):
Quickscan review.

scripts/export-users.py
SietsmaRJ: Simple TSV, I'll change it

marikaris (approved):

PR #482, opened 2023-09-12
Feat: refactor application.yml

clemens-tolboom:
Our `application.yml` contains several documents includes by using profiles.

We concluded we have only 2 modes for running the application so why not invoke using `spring.config.additional-location` command line arguments?

- [x] Decide which/how many files we need?
- [x] Can they reside in `armadillo/src/main/resources/` (simple include)
  - That is part of the jar so probably a bad location
  - moved into `armadillo/examples/` 
- [x] Is it user specific `application-security-oidc.yml.dist` (user provided values)
  - editor does not like `.yml.dist` so kept `.yml`
- [x] merging example files is more difficult then include multiple
- [x] remove/cleanup code regarding spring profiles
  - still puzzled about 
- [ ] has impact on the installer
- [x] document how users can enable debug

### Example

This includes basic auth and a custom user file

```
java -Dspring.profiles.active=development \
  -Dspring.config.additional-location=file:armadillo/src/main/resources/application-security-basic-auth.yml \
  -Dspring.config.additional-location=file:application-local.yml \
  -jar molgenis-armadillo-3.*.jar
```


armadillo/src/main/resources/application.yml
```diff
@@ -60,53 +60,3 @@ logging:
...
-      resourceserver:
-        jwt:
-          issuer-uri: 'https://auth.molgenis.org'
-        opaquetoken:
-          client-id: 'b396233b-cdb2-449e-ac5c-a0d28b38f791'
-  config:
-    activate:
-      on-profile: development
-
-logging:
```
clemens-tolboom, line 100: Should the application log by INFO and only DEBUG for development?

Logging is moved as a whole to basic auth yml

armadillo/src/main/resources/application-big-files.yml
```diff
@@ -0,0 +1,8 @@
+spring:
+  mvc:
+    async:
+      request-timeout: 36000000
+  servlet:
+    multipart:
+      max-file-size: 13000MB
+      max-request-size: 13000MB
```
clemens-tolboom, line 8: This came from @SietsmaRJ local config .. need to check for the docs whether we need this in here.

armadillo/src/main/resources/application-security-basic-auth.yml
```diff
@@ -0,0 +1,24 @@
...
+  level:
+    org.molgenis: DEBUG
+    org:
+      springframework:
+        boot:
+          actuate:
+            audit:
+              listener:
+                AuditListener: INFO
+          web:
```
clemens-tolboom, line 24: Strange dangling `web:`

armadillo/src/main/resources/application-security-basic-auth.yml
```diff
@@ -0,0 +1,24 @@
+spring:
+  # seem not to work
+  profiles:
+    active: development
+  security:
+    user:
+      name: administrator
```
clemens-tolboom, line 7: Reset to admin as we use this everywhere

armadillo/src/main/resources/application-security-oidc.yml
```diff
@@ -0,0 +1,19 @@
...
+  security:
+    oauth2:
+      client:
+        provider:
+          molgenis:
+            issuer-uri: 'https://auth.molgenis.org'
+        registration:
+          molgenis:
+            client-id: b396233b-cdb2-449e-ac5c-a0d28b38f791
+            client-secret: nRO_t1_cIpdzRzr-cWBeZg4ckBsMHmX2MlA9SaSg4P8
```
clemens-tolboom, line 13: Came from Robert ... not sure how to find these my self ... check docs

clemens-tolboom:
### Installer

```
diff --side-by-side scripts/install/conf/application.yml armadillo/src/main/resources/application.yml
```

clemens-tolboom:
We assume the is no need for `-Dspring.profiles.active=development` but I get

```
13:57:41.123 [main|] INFO o.m.a.ArmadilloServiceApplication - No active profile set, falling back to 1 default profile: "default"
...
Using generated security password: 2ac14745-29d4-4be0-b28e-a3eb18167c9a
```
using
```
java \    
  -Dspring.config.additional-location=file:armadillo/examples/application-security-basic-auth.yml \
  -Dspring.config.additional-location=file:application-local.yml \
  -jar build/libs/molgenis-armadillo-3.4.0-feat-refactor-applicaiton-yml-SNAPSHOT.jar
```

armadillo/src/main/resources/application.yml
mswertz: agreed, we should just document in README how to switch on debug

armadillo/src/main/resources/application-big-files.yml
mswertz: I think this should be moved to default application yaml?

clemens-tolboom: This is now in `application.yml` and `application.template.yml` for override which looks good to me.

armadillo/src/main/resources/application.yml
clemens-tolboom: Is is not in the README afaik but in `application.template.yml` IMHO non blocking

clemens-tolboom:
Not sure how to test the installer ... there is `scripts/install/README.md`

marikaris (approved):

PR #484, opened 2023-09-15
feat: user import script

SietsmaRJ:
- Added python3 script to import cohorts and users into an armadillo 3 server
- Synchronized pipfile and pipfile.lock with #480 
- Added utilitairy script to reset a localhost projects and users back to default, according to user supplied "default user"


marikaris:
[M10354 As armadillo server maintainer I want to import users from armadillo 2 in armadillo 3](https://trello.com/c/0f4ff9CM/1087-m10354-as-armadillo-server-maintainer-i-want-to-import-users-from-armadillo-2-in-armadillo-3)

scripts/reset-localhost-projects.py
```diff
@@ -0,0 +1,114 @@
...
+class CommandLineParser:
+    def __init__(self):
+        self.parser = CommandLineInterface()
+
+    def parse(self):
+        su = self._parse_su()
+        return su
+
+    def _parse_su(self):
+        return self.parser.get_argument("user")
```
clemens-tolboom, line 74: - What means `su`? Maybe expand in full word(s)?
- `_parse_su` is private and only called from `parse`? ... then you could fold code into `parse` itself no?

scripts/import-users.py
```diff
@@ -0,0 +1,410 @@
...
+        for cohort_name, cohort_users in cohort_data.items():
+            # Skip users that are not categorized into a cohort
+            if cohort_name == "uncategorized-users":
+                continue
+            self._obtain_su(cohort_users)
+            self._put_project(name=cohort_name, users=cohort_users['email'].values.tolist())
+        self._process_uncategorized_users(cohort_data['uncategorized-users'])
+        self._add_admin_rights()
+
+    @staticmethod
```
clemens-tolboom, line 130: Why static ... it is called non-static no? Are the other @static* similar needed?

scripts/import-users.py
```diff
@@ -0,0 +1,410 @@
...
+        # First, break off file extension
+        filename = filename.replace(".tsv.gz", "")
+        filename = filename.replace(".tsv", "")
+        # Then create an intermediate with only valid characters
+        name = "".join(
+            [c for c in filename.replace("-", "") if c.isalnum() or c == "_"]
+        )
+        # Then clean up starting spaces
+        name = re.sub(r"^_*", "", name)
+        name = re.sub(r"_+", "-", name)
```
clemens-tolboom, line 69: You can replace these with `.strip()` after breaking file extension. https://docs.python.org/3/library/stdtypes.html?highlight=strip#str.strip

scripts/reset-localhost-projects.py
clemens-tolboom: Hmm ... `su` is super user .. or is this users with admin role? (I myself thing linux `su`)

clemens-tolboom (commented):
quick scan review

scripts/import-users.py
SietsmaRJ: As discussed: staticmethods in python are methods that do not really use anything of the class they are defined in, also pycharm complains about methods that do not use any "self" in their use (init variables, other methods etc.)

scripts/import-users.py
SietsmaRJ: Spaces was incorrect in the comment line, it's to strip starting underscores. The second regex is to replace 1 or more underscores with a single - line that is allowed in AWS bucket naming scheme.

scripts/reset-localhost-projects.py
SietsmaRJ: It was meant as superuser yes, changed for clarification

clemens-tolboom:
I do miss additions to https://github.com/molgenis/molgenis-service-armadillo#how-to-import-data-from-armadillo-2 when to run the script(s).

And maybe a `README.md` in `scripts` would help for developers I guess.

erikzwart (changes requested):
I miss a minimal example of how to run this script, also some small introduction what this script supposes to do and what the requirements are. 

From what I understood a separate script will export users out of fusion-auth (or minio) and specifies to which cohort they belong. This TSV file can then be shared with the cohort (only share cohort relevant user data). The cohort manager can run this script (import-users.py) and import the users into Armadillo (and also the projects with the users)?

The minimal example and explanation should be added to the readme: https://github.com/molgenis/molgenis-service-armadillo/blob/master/README.md
Most likely in the FAQ after How to import data from Armadillo 2?

clemens-tolboom:
The documentation is moved to https://github.com/molgenis/molgenis-service-armadillo/blob/master/docs/migration-guide-2-to-3.md by PR #509

mswertz (approved):
looks good. Given approval of Erik, Mariska or Clemens you can merge.

clemens-tolboom:
I cannot run `reset-localhost-projects.py` as I get

```zsh
python3.8 reset-localhost-projects.py
Traceback (most recent call last):
  File "reset-localhost-projects.py", line 5, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

### MacOS

I did

```
brew install python@3.8 

python3.8 -m venv old
source ~/venv/old/bin/activate

cd .../molgenis-service-armadillo/scripts

pip3.8 install Pipfile
python3.8 reset-localhost-projects.py
```

SietsmaRJ:
@clemens-tolboom The documentation for installation of all python scrips inside `scrips` is changed in this PR, so that's why the github pages is not updated yet: https://github.com/molgenis/molgenis-service-armadillo/pull/484/files#diff-1c2f973391c26ad23290e5ef455050be3689216e3337b0459ef5195470ab5f98R12-R22

clemens-tolboom:
I reviewed last change but never submit my comment.

So I approve changes but @erikzwart should check how to the installation work out.

erikzwart:
I followed the instructions on docs/upgrade-2-3.md, 'Install with Python virtual environment'
cd scripts

Below the export of the users from the auth server (not the Armadillo what is a bit confusion reading the docs), it gives warnings but does export ALL data. Not a particular cohort which is not specified, so we should note that this script is only  used by molgenis admins.

(venv): python export-users.py -f https://AUTHSERVER.org -o export
export-users.py:99: UserWarning: There are 12 cohorts without users present: [PROD] Armadillo 3 - KANC, [PROD] Armadillo 3 - OPUS, [ACC] Armadillo 3 - Gen-R, [PROD] Armadillo 3 - Gen-R, [PROD] Armadillo 3 - Recetox, [PROD] WIKI, [ TEST ] Armadillo 3 - IPEC locatie Roqua, [PROD] Armadillo 3 - IPEC Azure, [PROD] Armadillo - Raine, [PROD] Armadillo - Dutch Famine Birth Cohort, [PROD] Armadillo - PiccoliPiu, [TEST] Opal
  warnings.warn(

The folder produced does contain the data but not sure if it is correct for instance:
https://armadillo-demo.molgenis.net/ which has ~15 users compared with the data exported from the AUTH server
[DEMO]_Armadillo2_demo.tsv  which has 20 users.

erikzwart (approved):

Issue #488, opened 2023-09-21
bug: service script should add configuration instead of replace application.yml


clemens-tolboom:
We used the `application.template.yml` testing migration from v2 to v3

https://github.com/molgenis/molgenis-service-armadillo/blob/265bb22feb59b4349f25fc97ab4cf233fb06180b/scripts/install/armadillo-setup.sh#L122

should be change into `SPRING.CONFIG.ADDITIONAL-LOCATION` (what about the dash?)

DickPostma:
It looks like environment with ., wont work on systemd. 
I tested it with this config and that seems to work:
Environment=SPRING_CONFIG_ADDITIONAL_LOCATION=/etc/armadillo/application.yml

PR #479, opened 2023-08-22
chore: update documentation for armadillo 2 to 3 migration

marikaris:

PR #489, opened 2023-09-21
Make custom application.yml additional.

clemens-tolboom:
See #488

### DIFF
--- a/armadillo/src/main/resources/application.yml
+++ b/armadillo/src/main/resources/application.yml
@@ -31,6 +31,7 @@ spring:
       max-request-size: 1000MB
   application:
     name: armadillo
+  security:
     user:
       name: admin
       password: admin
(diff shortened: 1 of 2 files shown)


armadillo/src/main/resources/application.yml
```diff
@@ -31,7 +31,6 @@ spring:
       max-request-size: 1000MB
   application:
     name: armadillo
-  security:
```
mswertz, line 34: this doesn't make sense to me?

DickPostma (approved):
LGTM

armadillo/src/main/resources/application.yml
clemens-tolboom: The whole file seems too much somehow. Logging is mixed with effort for PR #479

clemens-tolboom: Restored `security:`

PR #517, opened 2023-10-13
Bug/404 on whitelist

clemens-tolboom:
~~Test is incomplete as it needs a proper context.~~ removed test (fix later)


DickPostma (approved):
LGTM

PR #514, opened 2023-10-10
Add isRock variable to profiles

clemens-tolboom:
- [ ] Java needs test for existing profiles
- [x] Vue does not? render boolean
- [ ] Needs PR #512 to add tests
- [ ] Making it show as checkbox (instead of `true|false` text) does not show [x]
- [ ] Is `isRock` enough to remove 'Not a Rock server' code (flow)?

PR #518, opened 2023-10-13
Make rock selection better based on name or image

clemens-tolboom:
Inspired by #514 this PR makes it a little better and exposes the correct port from the container.

This is probably a better solution then go through the UI/Server stack as #514 tried to do since we plan to drop support for `rserve` soon.

A running server is either `rserve` or `rock` where `rserve` does not serve HTTP as curl shows depending on arguments

```zsg
curl --http0.9 --head http://localhost:6311
# curl: (8) Weird server reply

curl --no-http0.9 --head http://localhost:6311
# curl: (1) Received HTTP/0.9 when not allowed
```

while `rock` responds nicely with JSON

```zsh
curl --head http://localhost:6314 
```

>HTTP/1.1 200 OK
Date: Fri, 13 Oct 2023 11:55:01 GMT
Content-Type: application/json
X-Content-Type-Options: nosniff
X-XSS-Protection: 1; mode=block
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Pragma: no-cache
Expires: 0
X-Frame-Options: DENY
Content-Length: 68

```zsh
curl http://localhost:6314 
```
> { "id": "rserver", "type": "rock", "cluster": "default","tags": [] }

## Test

The `scripts/release/release-test.R` uses the name 'rock' and ran as a whole testing rock images. Changing to another profile name say **roll** does not work as image is connected to string 'rock'

https://github.com/molgenis/molgenis-service-armadillo/blob/b4a959329929e74b3d07180f67f606d951553d82/scripts/release/release-test.R#L54-L55

image

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/profile/DockerService.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/profile/DockerService.java
@@ -108,7 +108,9 @@ public class DockerService {
     }
 
     // TODO: have full port mapping in config or a by R server type?
-    int imageExposed = profileConfig.getName().contains("rock") ? 8085 : 6311;
+    boolean isRock =
+        profileConfig.getName().contains("rock") || profileConfig.getImage().contains("rock");
+    int imageExposed = isRock ? 8085 : 6311;
     ExposedPort exposed = ExposedPort.tcp(imageExposed);
     Ports portBindings = new Ports();
     portBindings.bind(exposed, Ports.Binding.bindPort(profileConfig.getPort()));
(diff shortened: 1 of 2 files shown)


mswertz:
I think you might be right. Ideally we move to Rock asap, so lets see (a) what projects would not be able to move if we would release a 'rock only' release tomorrow and more importantly (b) try to understand how we can test that rock actually works with Armadillo => can we have a pilot analysis with for example help of Tim?

clemens-tolboom:
Superseded by #521 so closing

PR #532, opened 2023-11-01
Armadillo controller including autoload.

clemens-tolboom:
This is my take for an Armadillo controller to control through the Armadillo API the docker containers.

We can `start`, `stop`, `restart`, `status` with `*All` and `autoStart`.

Using `autoStart` the settings from the environment can be used.


scripts/ops/armadilloctl.zsh
```diff
@@ -0,0 +1,173 @@
...
+
+# FIXME: loads from local files
+if [[ -f ".env" ]]
+then
+  echo "Sourcing .env"
+  source ".env"
+else
+  echo "Sourcing dev.env"
+  source "dev.env"
+fi
```
clemens-tolboom, line 163: I'm not happy with this code as the file location of the `.env` should be discoverable. Feedback is welcome

clemens-tolboom: I think about not including the `.env` but just expect that the environment vars are set when running.

clemens-tolboom: Removed and documentation added. Ready for testing.

marikaris:
[Armadillo controller including autoload](https://trello.com/c/tQxzZTG9/1110-armadillo-controller-including-autoload)

scripts/ops/env.dist
```diff
@@ -0,0 +1,11 @@
+# copy this file to ie "/etc/armadillo/acc.env"
```
marikaris, line 1: Why do you suggest another path than in the `.README`? I suggest keeping it the same.

marikaris (commented):
Do we maybe want to add the .env files to the `.gitignore`? I'd suggest adding `scrips/ops/conf/armadillo.acc.env` to `.gitignore` (this goes for `scripts/release/.env` as well, so if you agree, maybe you could add this one as well)

scripts/ops/env.dist
```diff
@@ -0,0 +1,11 @@
...
+
+# FIXME: We could / should test for admin user in overload config file instead?
```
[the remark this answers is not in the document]
clemens-tolboom: That's a weird sentence ... when installing Armadillo admin user / password is already in `application.yml`. I was somehow referring the that file :-O

scripts/ops/env.dist
clemens-tolboom: Thanks ... done

clemens-tolboom:
Set to WIP as it should be `bash` instead of `zsh`

scripts/ops/env.dist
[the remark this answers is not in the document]
clemens-tolboom: I've removed it for now. Let first move this to old environments to test :)

scripts/ops/README.md
```diff
@@ -0,0 +1,40 @@
+# Armadillo for OPS
+
+To monitor and control Armadillo and its DataShield docker containers we need to use the Armadillo API.
+
+## Armadillo Controller
+
+With `armadilloctl` you can control the DataShield docker containers.
```
mswertz, line 7: who is the envisioned user for this? I mean why is this outside of the app?

clemens-tolboom: From hindsight a checkbox in Armadillo was probably enough ... the code was already written as investigation/spinoff from the Rock only PR.

marikaris (approved):

Issue #535, opened 2023-11-03
/projects/{project}/objects/{object} endpoint doesn't return responsebody on error


marikaris:
How to reproduce:
Call the endpoint with a non-existing combination of project/object:
`http://localhost:8080/storage/projects/lifecycle/objects/core%2Fnonrep.par`

Expected: 
404 with response body

Observed:
404 without body

marikaris:
![](https://github.trello.services/images/mini-trello-icon.png) [Object endpoint doesn't return responsebody on error](https://trello.com/c/Z1CG0yre/1107-object-endpoint-doesnt-return-responsebody-on-error)

PR #536, opened 2023-11-03
fix: #535 object endpoint doesnt return responsebody on error

marikaris:
Test using swagger. Login on `localhost:8080`, then go to the swagger page (http://localhost:8080/swagger-ui/index.html#/storage/downloadObject). 
Click "Try it out", Fill in a project and non existent object and click execute. 
See a 404 with a json response body, something like:
``` json
{
  "timestamp": "2023-11-03T12:44:37.901+00:00",
  "status": 404,
  "error": "Not Found",
  "message": "Project 'project' does not exist",
  "path": "/storage/projects/project/objects/object"
}
```
Before the body used to be empty.


marikaris:
[Object endpoint doesn't return responsebody on error](https://trello.com/c/Z1CG0yre/1107-object-endpoint-doesnt-return-responsebody-on-error)

clemens-tolboom (approved):
LGTM ... tested and found expected JSON

PR #555, opened 2023-11-16
docs: add documentation for running armadillo using apache

marikaris:
...instead of nginx. We don't recommend or support this, but if anyone wants to use it, they can at their own risk using this documentation.


DickPostma (approved):
LGTM

erikzwart (approved):

PR #561, opened 2023-11-29
Require an image version like :latest

### LINKED ISSUE #524: Default profile container version not specified
When installing Armadillo fresh, the default profile does not mention a specific version (nor latest). This results in 22,11GB of additional images which (might) not be nessecairy?

clemens-tolboom:
Fixes #524

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -1,4 +1,14 @@
-# Vue 3 + Vite
+# Armadillo UI
+
+## Test as a UI developer
+
+- `cd ui/`
+- make sure you have `yarn` installed.
+- `yarn test --watch`
+- change one of `src/**`
+- change corresponding `tests/unit/**`
+
+## How this project was generated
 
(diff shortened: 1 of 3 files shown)


ui/README.md
```diff
@@ -1,4 +1,14 @@
-# Vue 3 + Vite
+# Armadillo UI
+
+## Test as a UI developer
+
+- `cd ui/`
+- make sure you have `yarn` installed.
+- `yarn test --watch`
+- change one of `src/**`
+- change corresponding `tests/unit/**`
```
SietsmaRJ, line 9: I know what you mean by this, but it can be preserved as confusing at first... Perhaps it's better to state the last points as "do your developer things but also remember to write and/or update the tests" (perhaps different wording)

SietsmaRJ (changes requested):
In `tests/unit/views/Profiles.spec.ts`, line 82, in the `singleTestData` variable, there is still an unmarked image tag.

ui/README.md
clemens-tolboom: @SietsmaRJ please check the rewrite

SietsmaRJ (approved):

Issue #574, opened 2023-12-07
Master build artifacts are not tagged and named correctly


clemens-tolboom:

clemens-tolboom:
In effort PR #575 the artifacts of https://github.com/molgenis/molgenis-service-armadillo/tags are not as expected.

- tag v4.0.0 is not created

clemens-tolboom:
Running

```
./gradlew assemble
```
did not create the expected versions either

```
ls build/distributions/
molgenis-armadillo-3.4.0-SNAPSHOT-7e878d2e.tar  molgenis-armadillo-boot-3.4.0-SNAPSHOT-7e878d2e.tar
molgenis-armadillo-3.4.0-SNAPSHOT-7e878d2e.zip  molgenis-armadillo-boot-3.4.0-SNAPSHOT-7e878d2e.zip
```

marikaris:
will look into when fixing master build, closing for now

PR #576, opened 2023-12-08
Make sure release publish to GitHub and Docker

clemens-tolboom:
As reported in #574 the artefact where wrongful named (and still not published on GitHub)

- [x] check for NYX version number (which depends on correct BREAKING) ... is now documented
- [ ] check for release on GitHub with fixed issues from #574
- [x] publish on GitHub done by 33c29b79bf1632e91c74e279379452de4b7e72db

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -175,7 +175,9 @@ jobs:
           path: build/artifacts/armadillo-compose.zip
 
       - run:
-        ./gradlew release
+        name: Release
+        command: |
+          ./gradlew release
 
 workflows:
   build-deploy:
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -173,7 +173,14 @@ jobs:
 
       - store_artifacts:
           path: build/artifacts/armadillo-compose.zip
+          destination: armadillo-compose.zip
```
mswertz, line 176: include version in file name?

clemens-tolboom: Now we have a first stab @ v4.0.0 I think and will test we do not have to version the CircleCI artifacts but need to transform nyx release by *adding* the version.

clemens-tolboom:
Processed this PR into master commits ... let's see what happens there :-O

Issue #579, opened 2023-12-12
swagger is slow on some endpoints making it unusable


clemens-tolboom:
Browser goes to 100% when ie opening an endpoint

## Slow on

- http://localhost:8080/swagger-ui/index.html#/DataSHIELD/loadTable

clemens-tolboom:
Caused by some complex regex patterns

```java
  public static final String SYMBOL_CSV_RE = "\\p{Alnum}[\\w.]*+(,\\p{Alnum}[\\w.]*+)*+";
...
`@Valid @Pattern(regexp = TABLE_RESOURCE_REGEX)`
```

clemens-tolboom:
This should either solved upgrading swagger or remove the java validation

marikaris:
Think we removed the validation that broke this

PR #590, opened 2023-12-14
chore: workaround issue #579 (swagger performance)

clemens-tolboom:


marikaris (changes requested):
So what I was afraid of, is actually true. This breaks the whole validation of this endpoint and does not just affect swagger, on this PR:
``` R
library(httr)
library(DSMolgenisArmadillo)

.handle_request_error <- function(response) {
  if (response$status_code == 401) {
    stop("Unauthorized", call. = FALSE)
  } else if (response$status_code == 403) {
    stop("Forbidden", call. = FALSE)
  } else if (response$status_code == 404 || response$status_code == 409) {
    stop(
      httr::content(response, as = "parsed", encoding = "UTF-8")$message,
      call. = FALSE
    )
  } else if (response$status_code == 500) {
    stop(
      paste0(
        "Internal server error: ",
        httr::content(response, as = "parsed", encoding = "UTF-8")$message
      ),
      call. = FALSE
    )
  } else if (response$status_code >= 300) {
    stop(
      paste0(
        "Something went wrong (",
        response$status_code,
        "): ",
        httr::content(response, as = "parsed", encoding = "UTF-8")$message
      ),
      call. = FALSE
    )
  }
}


armadillo_url <- "http://localhost:8080"
token <- armadillo.get_token(armadillo_url)
auth_header <- c("Authorization" = paste0("Bearer ", token))
builder$append(
  server = "armadillo",
  url = armadillo_url,
  token = token,
  driver = "ArmadilloDriver",
  profile = "xenon",
)

login_data <- builder$build()
conns <- DSI::datashield.login(logins = login_data, assign = FALSE)
query <- list(table = "lifecycle/core/nonrep", symbol = "core_n##(*@onrep", async = TRUE)
variables <- c("cohoD*#rt_id", "sex")
query$variables <- paste(unlist(variables), collapse = ",")

response <- httr::POST(
  handle = handle(armadillo_url),
  path = "/load-table",
  query = query,
  config = httr::add_headers(auth_header)
)
.handle_request_error(response)
```
--> Nothing, no error, but on the master, I get:
```
Error: Internal server error: loadTable.variables: must match "\p{Alnum}[\w.]*+(,\p{Alnum}[\w.]*+)*+", loadTable.symbol: must match "\p{Alnum}[\w.]*"
```
And then continue with this:
``` R
datashield.assign.table(conns, "my_sy*(&(mbol", "lifecycle/core/nonrep", variables = c("cohort_id","sex","agebirth_m_y", "br*(&^eastfed_any", "breastfed_ever", "eusilc_income"))
```
This succeeds:
```
Assigned all table (my_sy*(&(mbol <- ...) [============================================] 100% / 2s
``` 
Until, I ask for the colnames:
``` R
> ds.colnames("my_sy*(&(mbol", datasources = conns)
  Aggregated (exists("my_sy*(&(mbol")) [=======================>-------------------------]  50% /14sError: There are some DataSHIELD errors, list them with datashield.errors()
```
This one also fails on the assign already on the master. 



So in conclusion, although this fixes swagger, I think we should look into alternatives for this fix.

clemens-tolboom:
I found this **Exploiting Input Sanitization for Regex Denial of Service
** PDF https://docs.lib.purdue.edu/cgi/viewcontent.cgi?article=1179&context=ecepubs 

TLDR "Non terminated regex"

Is ours like that?

clemens-tolboom:
We could replace

```java
  public static final String SYMBOL_RE = "\\p{Alnum}[\\w.]*";
```

with a length limited version ... and why not from start to end validate? ChatGPT choose 39

```java
  public static final String SYMBOL_RE = "^\\p{Alnum}(?:[\\w.]{0,39})$"
```

For the more complex

```java
  public static final String SYMBOL_CSV_RE = "\\p{Alnum}[\\w.]*+(,\\p{Alnum}[\\w.]*+)*+";
```

I got a lookahead version

```regex
"^(?=.{1,40}$)\\p{Alnum}[\\w.]*+(,\\p{Alnum}[\\w.]*+)*+"
```

clemens-tolboom:
Not sure I understand the problem exactly ... tested a little on some of the endpoint of `./armadillo/src/main/java/org/molgenis/armadillo/controller/DataController.java`

```java
//  public static final String SYMBOL_RE = "\\p{Alnum}[\\w.]*";
  public static final String SYMBOL_RE = "^\\p{Alnum}(?:[\\w.]{0,39})$";
//  public static final String SYMBOL_CSV_RE = "\\p{Alnum}[\\w.]*+(,\\p{Alnum}[\\w.]*+)*+";
  public static final String SYMBOL_CSV_RE = "^(?=.{1,40}$)\\p{Alnum}[\\w.]*+(,\\p{Alnum}[\\w.]*+)*+";
```

marikaris:
We have more important issues right now and don't have time to spend on this. Will investigate and look at the comments in here when we have more time.

PR #592, opened 2023-12-15
fix: #591 loadTable returns tibble for arrow 13+

### LINKED ISSUE #591: loadTable function returns tibble for arrow 13 and up
How to reproduce:
``` R
library(DSMolgenisArmadillo)

armadillo_url <- "http://localhost:8080"
token <- armadillo.get_token(armadillo_url)
builder$append(
  server = "armadillo",
  url = armadillo_url,
  token = token,
  driver = "ArmadilloDriver",
  profile = "xenon",
)

login_data <- builder$build()
conns <- DSI::datashield.login(logins = login_data, assign = FALSE)
datashield.assign.table(conns, "my_symbol", "lifecycle/core/nonrep", variables = c("cohort_id","sex","agebirth_m_y", "breastfed_any", "breastfed_ever", "eusilc_income"))

ds.class(x = "my_symbol", datasources = conns)
```
Expected:
```
[1] "data.frame"
```

Observed:
```
[1] "tbl_df"     "tbl"        "data.frame"
```

marikaris:
instead of dataframe

closes #591

How to test: 
Execute code in https://github.com/molgenis/molgenis-service-armadillo/issues/591 and see that the type now equals dataframe instead of tibble.

### DIFF
--- a/r/src/main/java/org/molgenis/r/service/RExecutorServiceImpl.java
+++ b/r/src/main/java/org/molgenis/r/service/RExecutorServiceImpl.java
@@ -79,7 +79,7 @@ public class RExecutorServiceImpl implements RExecutorService {
       if (variables.isEmpty()) {
         execute(
             format(
-                "is.null(base::assign('%s', value={arrow::read_parquet('%s')}))",
+                "is.null(base::assign('%s', value={data.frame(arrow::read_parquet('%s', as_data_frame = FALSE))}))",
                 symbol, rFileName),
             connection);
       } else {
@@ -89,7 +89,7 @@ public class RExecutorServiceImpl implements RExecutorService {
                 + ")";
         execute(
             format(
-                "is.null(base::assign('%s', value={arrow::read_parquet('%s', col_select = %s)}))",
+                "is.null(base::assign('%s', value={data.frame(arrow::read_parquet('%s', as_data_frame = FALSE, col_select = %s))}))",
                 symbol, rFileName, colSelect),
             connection);
       }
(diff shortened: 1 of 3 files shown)


r/src/main/java/org/molgenis/r/service/RExecutorServiceImpl.java
```diff
@@ -79,7 +79,7 @@ public void loadTable(
       if (variables.isEmpty()) {
         execute(
             format(
-                "is.null(base::assign('%s', value={arrow::read_parquet('%s')}))",
+                "is.null(base::assign('%s', value={data.frame(arrow::read_parquet('%s', as_data_frame = FALSE))}))",
```
mswertz, line 82: ah, this makes sense actually

mswertz (approved):
Is there a way to add a test for this somehow in the release testing or something like that?

r/src/main/java/org/molgenis/r/service/RExecutorServiceImpl.java
marikaris: Haha, well, I thought it didn't make sense, `as_data_frame = FALSE` combined with `data.frame` is total madness, but it works.

marikaris:
> Is there a way to add a test for this somehow in the release testing or something like that?

Yes, it has to be tested for the researcher user (=OIDC) though, so it won't run in the CI tests for now. It runs fine locally: 
Screenshot 2023-12-15 at 10 35 09

PR #522, opened 2023-10-16
feat: rock only 519 

clemens-tolboom:
## Provides

- [x] rock only images for
  - [x] `release-test.R`
  - [x] `application.yml` default profile

## Needs

- [x] remove java code regarding 'rserver'
- [x] remove java code selecting correct service 'rock' or 'rserve'
- [x] 3 tests are disabled .. needs review how to fix (We switched to Rock which were not tests previously so I tried to fix those but failed.)
- [x] fix docker checks not trying to connect when server is not ready?

PR #595, opened 2024-01-08
Do same as PR 552

clemens-tolboom:
PR #552 test run fails when on local but removing `yarn.lock` locally (this PR) makes test succeed.

Running tests again locally with generated `ui/yarn.lock` file and removed `ui/node_modules` test fails.


clemens-tolboom:
Shelling into without a yarn.lock file tests success

Next we have `ui/node_modules` + `ui/yarn.lock`

Removing `ui/yarn.lock` and running `./gradlew clean :ui:test` succeeds

```bash
# root@7b141a75677e:~/repo# 
rm ui/yarn.lock 
./gradlew clean :ui:test
```

Removing the `ui/node_modules` gives the wrong modules

```bash
# root@7b141a75677e:~/repo# 
rm -r ui/node_modules 
./gradlew clean :ui:test
```

So when **building** `ui/yarn.lock` based on `ui/package.json` we get a different `ui/node_modules` tree then when **using** the `ui/yarn.lock` to fetch the `ui/node_modules` tree

clemens-tolboom:
LMGTFY: "running test using yarn fails second time"

BING.CHAT: "My `node_module` directory is different from the initial version from `package.json` after `yarn.lock` is generated"

clemens-tolboom:
Obsolete as #522 is now fixed

PR #603, opened 2024-01-15
bug: always do prerelease

clemens-tolboom:


clemens-tolboom:
Merged hoping to see version 4.0.3 as a pre release.

PR #609, opened 2024-01-22
Feat/make update testable

clemens-tolboom:
Changed the script:

- added echo's to see what vars/url are found and used
- added `mode=dev` or `non dev` to run `systemctl`
- added argument to download `pre-release` versions or older then `4.0.0.` versions
- auto install is possible (default:n)
- added version compare

## Test locally

- checkout repo
- `./gradlew run`
- `cd scripts/install`
- `./armadillo-check-update.sh`
- `./armadillo-check-update.sh 3.3.0`

## Test on server

- change mode to prd
- check armadillo path for valid and application downloads
- check value of auto install
- `./armadillo-check-update.sh`
- `./armadillo-check-update.sh 3.3.0`

### DIFF
--- a/scripts/install/armadillo-check-update.sh
+++ b/scripts/install/armadillo-check-update.sh
@@ -4,6 +4,8 @@
 # We assume all version are 3 numbers x.y.z
 #
 # Giving an argument that must be a version which forces this version to download.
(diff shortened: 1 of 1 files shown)


clemens-tolboom:
Discussed with @DickPostma ... is GTG

PR #615, opened 2024-01-24
feat: use actuator/metrics for data points

clemens-tolboom:
### Current

image

### Previous

image

### Other Actuator

This makes discussion easier as the links are available.

image

We can use de standard items and build custom

- [x] use standard
- [x] Through new issue ~~make custom rock items available by their API like `rserve.processes.current` already there~~
- [x] make UI table
- [ ] Add tests
- [ ] cleanup UI
- [ ] Align header same as Projects 
- [ ] Add search to filter on
  - image
- [x] Details clutters the table feel
  - image
  - [ ] Move Details in column as popout?
- [ ] odd/even on path difference .. detail above has 2 items on same path
- [ ] Bytes are rounded up to 2 digits precision B,KB,MB,GM,TB,EB
  - [ ] is similar nice to have for times ie uptime in days?
image

---

# Actuator
### By names

```bash
curl --silent --user admin:admin http://localhost:8080/actuator/metrics | json_pp
{
   "names" : [
      "application.ready.time",
      "application.started.time",
      "disk.free",
      "disk.total",
      "http.server.requests",
      "http.server.requests.active",
      "jvm.buffer.count",
      "jvm.buffer.memory.used",
...
```
#### fetch by name
```bash
# file: actuator-metrics
json=$(curl --silent --user admin:admin http://localhost:8080/actuator/metrics)

names=$(echo "$json" | jq -r '.names[]')
```bash
for name in $names; do
   echo "Processing $name"
   # Add your processing code here
done
```
```

```
### Groups

```bash
curl --silent --user admin:admin http://localhost:8080/actuator/metrics|
json_pp | grep '\.' | cut -d . -f 1 | uniq -c

   2       "application
   2       "disk
   2       "http
  22       "jvm
   1       "logback
   5       "process
   1       "rserve
  42       "spring
   3       "system
   6       "tomcat
```

### disk.free

```bash
curl --silent --user admin:admin http://localhost:8080/actuator/metrics/disk.free | json_pp

{
   "availableTags" : [
      {
         "tag" : "path",
         "values" : [
            "/Users/clemens/Documents/GitHub/molgenis-service-armadillo/."
         ]
      }
   ],
   "baseUnit" : "bytes",
   "description" : "Usable space for path",
   "measurements" : [
      {
         "statistic" : "VALUE",
         "value" : 269983137792
      }
   ],
   "name" : "disk.free"
}
```


clemens-tolboom:
Weird error while testing on `rserve.processes.current` ... is that an actuator item?

```
15:45:45.103 [http-nio-8080-exec-4|] WARN  o.m.r.RServerConnectionFactory - Container for 'http://localhost:6312'  is down
15:45:45.118 [http-nio-8080-exec-4|] WARN  i.m.c.i.internal.DefaultGauge - Failed to apply the value function for the gauge 'rserve.processes.current'. Note that subsequent logs will be logged at debug level.
org.molgenis.r.exceptions.ConnectionCreationFailedException: org.molgenis.r.rock.RockServerException: Failure when opening a Rock R session
        at org.molgenis.r.rock.RockConnectionFactory.tryCreateConnection(RockConnectionFactory.java:39)
        at org.molgenis.r.RServerConnectionFactory.tryCreateConnection(RServerConnectionFactory.java:66)
        at org.molgenis.armadillo.info.RProcessEndpoint.connect(RProcessEndpoint.java:58)
...
```

armadillo/src/main/java/org/molgenis/armadillo/info/FileMetrics.java
```diff
@@ -0,0 +1,45 @@
...
+
+import io.micrometer.common.lang.NonNull;
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.MeterBinder;
+import java.io.File;
+import org.springframework.stereotype.Component;
+
+@Component
+public class FileMetrics implements MeterBinder {
```
marikaris, line 11: Could you add a test for this?

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+  [key: string]: any;
+}
+/**
+ * Fetch all metric values on one go using the list.
+ */
+export async function getMetricsAll() {
+  try {
+    const metrics = await getMetrics();
+    const paths = Object.keys(metrics);
+    // FIXME: path.forEach is more appropiate
```
marikaris, line 116: Could we fix this before merge?

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+
+/**
+ * Get list of metrics.
+ */
+async function getMetrics() {
+  const path = "/actuator/metrics";
+  return await get(path)
+    .then((data) => {
+      // Check if the data has 'names' property
+      if (data.hasOwnProperty("names")) {
```
marikaris, line 140: Why do we do we alter the output here and don't we create the right output in the backend? (just curious, you probably have a good reason)

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+
+        // console.log(tree);
+        return JSON.parse(JSON.stringify(tree["_bare"]));
+      } else {
+        console.log("No names found in the data");
+        return {};
+      }
+    })
+    .catch((error) => {
+      console.error(`Error fetching ${path}`, error);
```
marikaris, line 157: we want to be able to see errors in the ui, so they need to be passed through the api call

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+      return {};
+    });
+}
+
+async function getActuators() {
+  return await get("/actuator");
```
marikaris, line 163: return the get, not the await, so we can show loading spinner

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+const loadActuator = async () => {
+  let result = (await getActuator())["_links"];
+  let list = [];
+  for (let key in result) {
+    const item = result[key];
+    item["key"] = key;
+    list.push(result[key]);
+  }
+  actuator.value = list;
+  console.log(JSON.stringify(actuator.value, null, 2));
```
marikaris, line 19: remove console.log

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+import ActuatorItem from "./ActuatorItem.vue";
+import SearchBar from "@/components/SearchBar.vue";
+
+const actuator = ref({});
+const metrics = ref({});
+
+const loadActuator = async () => {
+  let result = (await getActuator())["_links"];
+  let list = [];
+  for (let key in result) {
```
marikaris, line 13: Not entirely sure what you're trying to do here. Feels duplicate somehow. You're trying to add the key as a value to the item? Why do we need this? Can't we solve this when we're processing the data?

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+
+  // preload search values
+  filteredLines();
+};
+
+loadMetrics();
+loadActuator();
+function downloadJSON(json, filename: string) {
+  const cleanedUp = removeFields(json);
+  var dataStr =
```
marikaris, line 32: Please use `const` and `let` instead of `var`

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+const SEARCH_TEXT_FIELDS = "searchWords";
+
+/**
+ * Filter metrics on search value
+ *
+ * We add:
+ * - string search field for matching
+ * - booelan display field for storing matched
+ */
+function filteredLines() {
```
marikaris, line 60: please write a test for this

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+  const filterOn: string = filterValue.value.toLowerCase();
+  for (let [key, value] of Object.entries(metrics.value)) {
+    if (!value[SEARCH_TEXT_FIELDS]) {
+      // TODO: drop keys before stringify?
```
marikaris, line 64: No TODO's in prod code.

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+      value[SEARCH_TEXT_FIELDS] = JSON.stringify(value).toLowerCase();
+    }
+    const searchWords: string = value[SEARCH_TEXT_FIELDS];
+    value[FIELD_DISPLAY] = filterOn === "" || searchWords.includes(filterOn);
```
marikaris, line 68: what does this `FIELD_DISPLAY` thing do?

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
...
+      <table>
+        <thead>
+          <tr>
+            <td>key</td>
+            <td>href</td>
+            <td>templated</td>
+          </tr>
+        </thead>
+        <tbody>
+          <tr v-for="item in actuator">
```
marikaris, line 127: please add key

ui/src/components/ActuatorItem.vue
```diff
@@ -0,0 +1,77 @@
...
+    type: Object,
+    required: true,
+  },
+});
+
+/**
+ * Convert given bytes to 2 digits precision round exponent version string.
+ * @param bytes number
+ */
+function convertBytes(bytes: number): string {
```
marikaris, line 17: this seems like a helper function to me, please move to helper/utils and add test

ui/src/components/ActuatorItem.vue
```diff
@@ -0,0 +1,77 @@
...
+  while (bytes >= 1024 && unitIndex < units.length - 1) {
+    bytes /= 1024;
+    unitIndex++;
+  }
+
+  return `${bytes.toFixed(2)} ${units[unitIndex]}`;
+}
+
+/*
+{
```
marikaris, line 30: Please remove uncommented code

ui/src/components/ActuatorItem.vue
```diff
@@ -0,0 +1,77 @@
...
+         "values": [
+            "org.molgenis.armadillo.ArmadilloServiceApplication"
+         ]
+      }
+   ]
+}
+*/
+</script>
+<template>
+  <tr v-if="data._display" v-for="(v, k) in data.measurements">
```
marikaris, line 52: please add key

ui/src/components/ActuatorItem.vue
```diff
@@ -0,0 +1,77 @@
...
+        <i v-if="data.description" class="bi bi-info-circle-fill"></i>
+      </span>
+    </td>
+    <td>{{ v.statistic }}</td>
+    <td v-if="data.baseUnit === 'bytes'">
+      {{ convertBytes(v.value) }}
+    </td>
+    <td v-else>{{ v.value }} {{ data.baseUnit }}</td>
+  </tr>
+  <tr v-if="data._display">
```
marikaris, line 66: Ah! This is why you use that display thingy, right? I would set this up a bit differently, like we did throughout the rest of the application. I would compute the filtered items or create a method, then only show those and in case there is no searchword, then filtered items should return the full list. If you do this, you don't have to worry about resetting things when the search is emptied. 
Example: https://github.com/molgenis/molgenis-service-armadillo/blob/23272d29bb66cc343e058587ad9dc5945897c99b/ui/src/views/Projects.vue#L310
I also changed it on the branch I created for this, so you could also look at that.

ui/src/views/Insight.vue
```diff
@@ -1,19 +1,50 @@
...
-    <div class="row">
-      <div class="col">
-        <div class="row">
-          <div class="col">
-            <h3>Files</h3>
-            <RemoteFiles />
-          </div>
-        </div>
+  <h2>Insight</h2>
+  <!-- FIXME: Cannot use tabs as it
```
marikaris, line 3: I don't think you should use tabs here. Tabs in tabs are generally confusing, but I think we already agreed upon a solution :)

ui/src/views/Insight.vue
```diff
@@ -1,19 +1,50 @@
...
 <script setup lang="ts">
+import Actuator from "@/components/Actuator.vue";
 import RemoteFiles from "@/components/RemoteFiles.vue";
+import { ref } from "vue";
+
+let currentTab = ref(0);
+const tabs = ["Files", "Server metrics"];
 </script>
+
+<style scoped>
```
marikaris, line 45: why is this there? don't think we're using the red border thingy, so we don't need the style?

ui/src/views/Users.vue
```diff
@@ -124,6 +124,7 @@ import SearchBar from "@/components/SearchBar.vue";
 import Table from "@/components/Table.vue";
 import InlineRowEdit from "@/components/InlineRowEdit.vue";
 import FeedbackMessage from "@/components/FeedbackMessage.vue";
+
 import { deleteUser, getUsers, putUser, getProjects } from "@/api/api";
```
marikaris, line 128: please remove whitespace

marikaris (changes requested):
As discussed. For now I'm not going to comment on the composition/options api. We will the refactor for that later. I do think we need tests and the typing should be fixed. Do you have the following plugins in your vsCode:
`TypeScript Vue Plugin` and `Vue Language Features (Volar)` ? I think it's the latter. I get error like these: 
Screenshot 2024-02-26 at 10 43 25
 :
Screenshot 2024-02-26 at 10 42 59
Screenshot 2024-02-26 at 10 43 07
Screenshot 2024-02-26 at 10 43 17

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+
+        // Process each name
+        data.names.forEach((name: string) => {
+          tree["_bare"][name] = {};
+        });
+
+        // console.log(tree);
+        return JSON.parse(JSON.stringify(tree["_bare"]));
+      } else {
+        console.log("No names found in the data");
```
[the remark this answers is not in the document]
clemens-tolboom: That was bad code indeed and is fixed now

ui/src/api/api.ts
clemens-tolboom: Fixed

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+  const path = `/actuator/metrics/${id}`;
+  let result = await get(path)
+    .then((data) => {
+      return JSON.parse(JSON.stringify(data));
+    })
+    .catch((error) => {
+      console.error(`Error fetching ${path}`, error);
+      return {};
+    });
+  return result;
```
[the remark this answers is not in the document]
clemens-tolboom: Fixed

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,139 @@
+<script setup lang="ts">
```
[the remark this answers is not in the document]
clemens-tolboom: Done

ui/src/components/Actuator.vue
clemens-tolboom: Done

ui/src/components/Actuator.vue
clemens-tolboom: Done

ui/src/components/Actuator.vue
clemens-tolboom: I disagree as a TODO can be needed later on in the project.

ui/src/components/Actuator.vue
clemens-tolboom: Is for next phase

ui/src/components/ActuatorItem.vue
clemens-tolboom: It's a one of AFAIK so local helper

ui/src/components/ActuatorItem.vue
clemens-tolboom: This is structure docs for discussing the template so keep it for now

mswertz (approved):
approve but please check if you can add smoketest for the '/actuator' and '/metrics'
I expect you would like to copy the test for the UserController

armadillo/src/main/java/org/molgenis/armadillo/info/FileMetrics.java
clemens-tolboom: will do in next iteration.

ui/src/api/api.ts
```diff
@@ -103,6 +103,85 @@ export async function getVersion() {
...
+      paths.map(async (path) => {
+        const response = await getMetric(path);
+        metrics[path] = response;
+        return { path: response };
+      })
+    );
+
+    return metrics;
+  } catch (error) {
+    console.error("Error:", error);
```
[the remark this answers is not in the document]
clemens-tolboom: Done

ui/src/api/api.ts
clemens-tolboom: Done

ui/src/components/Actuator.vue
clemens-tolboom: Gone

ui/src/components/ActuatorItem.vue
clemens-tolboom: Done

ui/src/components/ActuatorItem.vue
clemens-tolboom: We can check later in merged code

ui/src/views/Users.vue
clemens-tolboom: Removed

clemens-tolboom: I keep it as it separates libs from components

ui/src/views/Insight.vue
clemens-tolboom: Removed

ui/src/views/Insight.vue
clemens-tolboom: Will fix after merge

ui/src/components/Actuator.vue
clemens-tolboom: done

ui/src/components/Actuator.vue
clemens-tolboom: Adding the key into the item

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,174 @@
...
+const metrics = ref<Metrics>([]);
+const isLoading = ref<boolean>(true);
+
+const loadActuator = async () => {
+  let result = (await getActuator())["_links"];
+  let list = [];
+  for (let key in result) {
+    const item = result[key];
+    item["key"] = key;
+    list.push(result[key]);
```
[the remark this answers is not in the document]
clemens-tolboom: Code is different but I remember `api` failure must end in the UI not processed in api itself. I partly fixed that.

ui/src/components/Actuator.vue
```diff
@@ -0,0 +1,174 @@
+<template>
+  <div class="row">
+    <div class="col mt-3" v-if="isLoading">
+      <LoadingSpinner />
```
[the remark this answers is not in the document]
clemens-tolboom: I'm not sure how to add error within component ... will discuss after merge

ui/src/components/Actuator.vue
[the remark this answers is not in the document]
clemens-tolboom: Test is #701 which fails for some weird reasons ... will make separate PR

ui/src/components/Actuator.vue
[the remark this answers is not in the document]
clemens-tolboom: Is followup issue #702

PR #649, opened 2024-02-20
feat: m10506 views on tables

marikaris:
Fresh checkout from master with all commits from [feat/M10506-views ](https://github.com/molgenis/molgenis-service-armadillo/pull/563)


clemens-tolboom:
As we had two different errors ... creating and using links ... I wonder

- could it be a timing problem on CI ... what happens when sleeping a little say 10 seconds
- will it happen locally using `release-test.R` against armadillo-compose.zip (one of the artefacts from the CI

clemens-tolboom:
I think this is a CircleCI probleem which will eat lot of time

In the mean time we could merge this by excluding the failing test **only on CI** by
- exclude using .env `run_subview_on_dev` or `skip_subview_on_ci`
- not sure to make it a dev variable or CI one
- it should always run on DEV

marikaris:
> I think this is a CircleCI probleem which will eat lot of time
> 
> In the mean time we could merge this by excluding the failing test **only on CI** by
> 
> * exclude using .env `run_subview_on_dev` or `skip_subview_on_ci`
> * not sure to make it a dev variable or CI one
> * it should always run on DEV

Just thought of something. I removed the OIDC user in my .env file locally, then I get the same error:
`x Unable to create linked object 7uc21xy643/core-variables/nonrep from source: lifecycle/core/nonrep, status code: 405
Error: Unable to create linked object 7uc21xy643/core-variables/nonrep from source: lifecycle/core/nonrep, status code: 405
Execution halted
`

So I think the link doesn't work using the local user. Have to look into that.

clemens-tolboom (approved):
LGTM for now

PR #652, opened 2024-02-21
test: Test/dsmtl

timcadman:
Added dsMTL test. Note these operations take a long time to run so I've only tested one function from this package.

### DIFF
--- a/scripts/release/install_release_script_dependencies.R
+++ b/scripts/release/install_release_script_dependencies.R
@@ -55,6 +55,7 @@ warnings()
 
 library("devtools")
 install_github("datashield/dsBaseClient")
-install_github("datashield/dsMediationClient")
+install_github("https://github.com/transbioZI/dsMTLClient")
+
 #check if all packages are installed
 cli_alert_success("All packages are installed")
--- a/scripts/release/install_release_script_dependencies.R
+++ b/scripts/release/install_release_script_dependencies.R
@@ -55,7 +55,7 @@ warnings()
 
 library("devtools")
 install_github("datashield/dsBaseClient")
-install_github("datashield/dsMediationClient")
+install_github("datashield/dsMediationClient", ref = "0.0.3")
 install_github("https://github.com/transbioZI/dsMTLClient", ref = "0.9.9")
 
 #check if all packages are installed
--- a/scripts/release/install_release_script_dependencies.R
+++ b/scripts/release/install_release_script_dependencies.R
@@ -56,7 +56,7 @@ warnings()
 library("devtools")
 install_github("datashield/dsBaseClient")
 install_github("datashield/dsMediationClient")
-install_github("https://github.com/transbioZI/dsMTLClient")
+install_github("https://github.com/transbioZI/dsMTLClient", ref = "0.9.9")
 
 #check if all packages are installed
 cli_alert_success("All packages are installed")
--- a/scripts/release/install_release_script_dependencies.R
+++ b/scripts/release/install_release_script_dependencies.R
@@ -55,6 +55,7 @@ warnings()
 
 library("devtools")
 install_github("datashield/dsBaseClient")
+install_github("datashield/dsMediationClient")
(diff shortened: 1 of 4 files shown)


scripts/release/install_release_script_dependencies.R
```diff
@@ -55,6 +55,7 @@ warnings()
 
 library("devtools")
 install_github("datashield/dsBaseClient")
-install_github("datashield/dsMediationClient")
+install_github("https://github.com/transbioZI/dsMTLClient")
```
clemens-tolboom, line 58: Are there no CRAN versions of these? Guess not ... is it possible to add version argument for `install_github`

scripts/release/release-test.R
```diff
@@ -922,6 +967,17 @@ verify_ne_model_class()
 verify_ne_imp_class()
 verify_ne_lht_class()
 
+logindata_1 <- create_dsi_builder(server = "testserver1", url = armadillo_url, profile = profile, password = admin_pwd, token = token, table = sprintf("%s/2_1-core-1_0/nonrep", project1))
+logindata_2 <- create_dsi_builder(server = "testserver2", url = armadillo_url, profile = profile, password = admin_pwd, token = token, table = sprintf("%s/2_1-core-1_0/nonrep", project1))
+logindata <- rbind(logindata_1, logindata_2)
```
clemens-tolboom, line 972: Maybe document this setup? Nice way to test 2 servers.

I guess you should use two different tokens to make it more real world

scripts/release/install_release_script_dependencies.R
timcadman: No CRAN versions but I agree - I've added versions for both packages

scripts/release/release-test.R
timcadman: I agree two tokens would be more real world, but I've kept it like this to try to keep test efficient and reduce number of tokens required.

clemens-tolboom: I think it has test value for ie 2 sessions breaking apart

scripts/release/install_release_script_dependencies.R
clemens-tolboom: :)

scripts/release/release-test.R
timcadman: Ok done

timcadman:
> > Note these operations take a long time to run so I've only tested one function from this package.
> 
> What is long?
> 
> Checking with https://app.circleci.com/pipelines/github/molgenis/molgenis-service-armadillo it seems not an issue

About 3 minutes per function. Happy to add more if you think it's worth the length it will add?

scripts/release/release-test.R
timcadman: I've looked at this again, and whilst having two tokens is a nice suggestion we can't retrieve tokens in the CI (this was what caused it to fail). Therefore we would have to write two different test cases, one for the CI with the one token and one for running locally using two tokens combined with OIDC. So for now maybe we just keep the case with the one token that runs on both the CI and locally?

clemens-tolboom: CICD only runs test as admin as we cannot (yet?) connect to an OIDC account.

The section in the `README.md` could use some rewording.

```md
To prevent all the questions:

- copy the `dev.dist.env` to `.env`
- fill in some or all parts
- toggle the `interactive` to `n` or `y` to let the script wait for manual checks.
```

as it is not only useful  to prevent all questions.

Or we could extend documentation in the `dist.env`

Feel free to update that too in this PR

timcadman: Done

scripts/release/xenon-survival.R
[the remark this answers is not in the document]
timcadman: Oops yes, not sure how this got here ...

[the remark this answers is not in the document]
timcadman: Deleted

clemens-tolboom (approved):
LGTM

PR #659, opened 2024-02-22
fix(test): Added tests for dsSurvival

timcadman:
- I've separated out the tests in a separate script ('xenon-survival.R') which is then called within 'release-test.R'
- There's quite a lot of code repetition in 'xenon-survival.R', e.g.:

```
  if(surv_class$armadillo == "Surv"){
    cli_alert_success("ds.survival passed")
  } else{
    cli_alert_danger("ds.survival failed")
    exit_test("ds.survival did not create a serverside object with the expected class")
  }
```
Is repeated in a similar form in each test. I could break this into a separate function, but this may be at the cost of making it a bit less readable. Let me know what you'd prefer.


scripts/release/release-test.R
```diff
@@ -868,9 +879,9 @@ cli_alert_info(sprintf("Creating project [%s]", omics_project))
 armadillo.create_project(omics_project)
 rda_file_body <- upload_file(rda_dir)
 cli_alert_info(sprintf("Uploading resource file to %s into project [%s]", armadillo_url, omics_project))
-system.time({
-  post_resource_to_api(omics_project, token, auth_type, rda_file_body, "ewas", "gse66351_1.rda")
-})
+# system.time({
```
marikaris, line 882: Why did you comment this out?

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,109 @@
+create_survival_object <- function(){
```
marikaris, line 1: Ok, sorry for being nit-picky about this, but could you insert a space between the `)` and `{` ? (also in the rest of the code)

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,109 @@
...
+
+}
+
+verify_cox_output <- function(){
+
+  cox_output <- dsSurvivalClient::ds.coxph.SLMA(formula = 'surv_object~survival$age')
+  expected_names <- c("call", "fail", "na.action", "n", "loglik", "nevent", "coefficients",
+                      "conf.int", "logtest", "sctest", "rsq", "waldtest", "used.robust", "concordance")
+
+  if(identical(names(cox_output$armadillo), expected_names)){
```
marikaris, line 29: Here also add a space after `if` and before `{` (same below, and above)

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,109 @@
...
+    cli_alert_success("ds.coxph.SLMA passed")
+  } else{
```
marikaris, line 31: space between `else` and `{` (same below, and above)

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,109 @@
...
+
+  dsSurvivalClient::ds.Surv(
+    time='survival$time',
+    event = 'survival$status',
+    objectname='surv_object')
+
+}
+
+verify_survival_class <- function(){
+
```
marikaris, line 11: Remove empty line (functions can start with code, no need for empty line, same goes for the other functions)

marikaris (changes requested):
Put a couple of comments in. Minor things really (being neurotic about whitespaces mostly). 

Could you also add the testfiles in the `shared/lifecycle` folder? I think the tests will succeed if you do :)

scripts/release/release-test.R
timcadman: Ah that was a mistake. This step takes a long time to run so when testing the script I commented it out to speed things up - added back in :)

scripts/release/xenon-survival.R
timcadman: Done - I appreciate aesthetics so not nit-picky at all :)

scripts/release/xenon-survival.R
timcadman: Done

scripts/release/xenon-survival.R
timcadman: Done

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,103 @@
...
+    exit_test("ds.coxphSummary did not create a list with the expected elements")
+  }
+
+}
+
+library(dsSurvivalClient)
+
+logindata <- create_dsi_builder(url = armadillo_url, profile = profile, password = admin_pwd, token = token, table = sprintf("%s/survival/veteran", project1))
+cli_alert_info(sprintf("Login with profile [%s] and table: [%s/survival/veteran]", profile, project1))
+conns <- datashield.login(logins = logindata, symbol = "survival", assign = TRUE)
```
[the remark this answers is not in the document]
timcadman: I've tried to refactor a bit following these suggestions:

- I realised we don't need to log in again, we can just assign a table so I've removed lines 79 & 81 and added a line assigning.
- I've added a higher level function which calls the individual tests, and takes args for data_path, conns and project.
- xenon-survival.R now only contains function definitions, but does not call them.
- xenon-survival.R is sourced in the main script, and then the global function called.

docker/ci/Dockerfile
```diff
@@ -23,6 +23,8 @@ RUN /cicd/scripts/release/install_release_script_dependencies.R
 # Copy rest from of scripts
 COPY armadillo/scripts/release/* /cicd/scripts/release/
 
+COPY armadillo/scripts/release/xenon-survival.R /cicd/scripts/release/
+
```
clemens-tolboom, line 27: this should go

The error is in `docker/bin/prepare.bash`

But we need to put all extracted scripts into a folder say `lib` or `src` then include them from there

scripts/release/xenon-survival.R
```diff
@@ -0,0 +1,100 @@
...
+run_survival_tests <- function(project, data_path, conns) {
+  assign_survival_data(project = project, data_path = data_path, conns = conns)
+  create_survival_object()
+  verify_survival_class()
+  verify_cox_output()
+  verify_cox_phSLMAassign_class()
+  verify_cox_zphSLMA_object()
+  verify_cox_phsummary()
+
+}
```
clemens-tolboom, line 100: Needs whittling/newline

docker/ci/Dockerfile
clemens-tolboom: Extracted from `release-test.R` I meant

docker/bin/prepare.bash
```diff
@@ -63,6 +63,7 @@ LOG_DIR="$ARMADILLO_ROOT/log/"
 mkdir -p "$LOG_DIR" || exit 1
 
 cp "$PROJECT_DIR/scripts/release/release-test.R" "$BIN_DIR/" || exit 1
+cp -r "$PROJECT_DIR/scripts/release/release-test-cases" "$BIN_DIR/" || exit 1
```
clemens-tolboom, line 66: Nice dir name ... maybe make it shorter
- from`/release/release-test-cases`
- into `/release/test-cases`

That makes less names with `r` in it so easier to type expand in the terminal

scripts/release/release-test.R
```diff
@@ -1008,6 +1019,11 @@ verify_ne_model_class()
 verify_ne_imp_class()
 verify_ne_lht_class()
 
+cli_alert_info("Testing dsSurvival")
+list.files(".")
+source("/cicd/scripts/release/xenon-survival.R")
```
clemens-tolboom, line 1024: This is strange ... this should be a relative to `release-test.R` script include.

If it fails check sourcing from https://github.com/molgenis/molgenis-service-armadillo/blob/master/scripts/rock/rockctl/src/rockctl.R#L3-L16

marikaris (approved):

PR #664, opened 2024-02-27
Preview a branch

clemens-tolboom:
With this script we can preview a PR or other branch on a configured machine

```bash
ssh root@machine "~/auto_pre_release/preview-branch.bash BRANCH_NAME"
```

See README.md for more info


clemens-tolboom:
Discussed: this is not what we need ... we need similar as EMX2 is doing

PR #668, opened 2024-02-29
fix: add pager to insight files

clemens-tolboom:
## Tasks

- [x] ? how to make clear which does what
  - Page num:
    - negative value means from the end
    - 0 is first page, positive means from begin
  - Page size is number of lines from audit/log file
- [ ] ? make page size a dropdown 10, 20, 50, 100, 200, 500, 1000, All
- [ ] write tests
- [ ] is it performant enough?
- [ ] ?

## Current

image

## Previous

image

---

image


mswertz (approved):
I propose to merge. Meanwhile we can think on how the UI might be easier to understand between the 'paging' and the 'search'.

PR #666, opened 2024-02-27
java: Bump to java 19 fixes #663

clemens-tolboom:
On ubuntu server/desktop we cannot compile Armadillo without this version 19 as they have openjdk v19 installed.

Issue #663, opened 2024-02-27
Cannot compile on java version 19


clemens-tolboom:
Having version `19.0.2` 
```bash
java --version
openjdk 19.0.2 2023-01-17
OpenJDK Runtime Environment (build 19.0.2+7-Ubuntu-0ubuntu322.04)
OpenJDK 64-Bit Server VM (build 19.0.2+7-Ubuntu-0ubuntu322.04, mixed mode, sharing)
```

then running `./gradlew build` gives

```bash
./gradlew build
Downloading https://services.gradle.org/distributions/gradle-7.6.3-bin.zip
...........10%............20%...........30%............40%............50%...........60%............70%............80%...........90%............100%

Welcome to Gradle 7.6.3!

Here are the highlights of this release:
 - Added support for Java 19.
 - Introduced `--rerun` flag for individual task rerun.
 - Improved dependency block for test suites to be strongly typed.
 - Added a pluggable system for Java toolchains provisioning.

For more details see https://docs.gradle.org/7.6.3/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)
Project version: v4.2.0-SNAPSHOT.1

> Configure project :
Project previous version (nyx): v4.1.0
Corrected version checking for optional snapshot: 4.2.0-SNAPSHOT
Project version : 4.2.0-SNAPSHOT
Git version     : 37c6f102
Git name-ref    : remotes/origin/feat/metrics

FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':bootJar'.
> Could not resolve all task dependencies for configuration ':runtimeClasspath'.
   > Could not resolve project :armadillo.
     Required by:
         project :
      > No matching variant of project :armadillo was found. The consumer was configured to find a runtime of a library compatible with Java 17, packaged as a jar, preferably optimized for standard JVMs, and its dependencies declared externally but:
          - Variant 'apiElements' capability org.molgenis:armadillo:4.2.0-SNAPSHOT declares a library, packaged as a jar, and its dependencies declared externally:
              - Incompatible because this component declares an API of a component compatible with Java 19 and the consumer needed a runtime of a component compatible with Java 17
              - Other compatible attribute:
                  - Doesn't say anything about its target Java environment (preferred optimized for standard JVMs)
          - Variant 'coverageDataElementsForTest' capability org.molgenis:armadillo:4.2.0-SNAPSHOT:
[...]

PR #681, opened 2024-03-05
Build on 19.0.2_7-jdk-jammy

clemens-tolboom:
This is for reproduce build failing on Java 19 with Gradle asking for Java 17  #663 #666 

and testing upgrades to other Java versions.

### Failing CICD

I can open the build report root `build/reports/jacoco/jacocoMergedReport/html/index.html` using running

```bash
./gradlew jacocoMergedReport

open build/reports/jacoco/jacocoMergedReport/html/index.html
```

So is sonar calling the wrong function? We have

```groovy
   id "org.sonarqube" version "3.5.0.2730"
```

which is (if this is its repo) https://github.com/SonarSource/sonarqube/tags?after=3.6 is from **Mar 13, 2013**

While https://github.com/molgenis/molgenis-emx2/blob/master/build.gradle#L11 is using
```groovy
    id "org.sonarqube" version "4.4.1.3373"
```


clemens-tolboom:
Error is now 
> 'org.gradle.api.provider.Provider org.gradle.api.reporting.Report.getOutputLocation()'

Check with https://github.com/molgenis/molgenis-emx2/pull/2069/files#diff-cd279e89731dcd028c2a5741b1f8ba13cdd851ff187ade540479f55aa89dbedd seems ok apart from we list all report locations instead

clemens-tolboom:
Failure on Azure is probably cause by Java 17

We can upgrade to `/usr/lib/jvm/temurin-21-jdk-amd64/bin/java` to test Azure succeeds but do we want to upgrade for all?

PR #696, opened 2024-03-12
Delete application.template.yml

clemens-tolboom:


clemens-tolboom:
There was something going wrong using online editing ... can't remember what ... cannot remove this PR either :-(

Issue #704, opened 2024-03-14
Sonar code coverage are not generated


clemens-tolboom:
### How to Reproduce

- Visit https://github.com/molgenis/molgenis-service-armadillo
- Check the shields ![image](https://github.com/molgenis/molgenis-service-armadillo/assets/371014/78e2e9df-f94d-4675-b397-30313ed82c37)

### Expected behavior

- Coverage should have a non zero percentage
- Quality gate should not fail

### Observed behavior

- Coverage: 0%
- Quality gate: failed

clemens-tolboom:
PR #695 seems not to fix anything ... checking the CICD logs shows

> Task :jacocoTestReport SKIPPED

Trying to run locally

```
./gradlew :jacocoTestReport
# Does nothing
```

and

```
./gradlew jacocoTestReport
# runs test
```
ends with
```
...
RExecutorServiceImplTest > testLoadResource() PASSED
#
BUILD SUCCESSFUL in 36s
```

marikaris:
Fixed in #734

PR #709, opened 2024-03-18
Make java gradles similar

clemens-tolboom:


clemens-tolboom:
Wrong idea ... keep change in #704

PR #718, opened 2024-03-26
feat: simpler install

clemens-tolboom:
## how to test

- Visit the Users tab (will be chance)

image

## todo

- [ ] java: add endpoint to read/write property file in `data/system/config.properties`
- [ ] find place in UI for Config
  - overlay with cogwheel
- [ ] add tests


marikaris:
Will look into this again if we need it

PR #729, opened 2024-04-30
Feat: M10506 UI linkfile edit

marikaris:
NB: Look at this one after: #728 is merged (this one contains commits of that one, once that one's merged, it will all make sense again)

how to test:
- open project with linkfile (or create link)
- select existing linkfile
- see that the source is shown above the preview table
- click edit 
- edit variables
- save
- open linkfile again
- see variables changed

todo:
- [ ] added tests


mswertz (approved):
very nice. I only wish that after editing the link I would go back to the preview, but that is a very minor thing that when expensive we don't need now.

PR #733, opened 2024-05-03
fix: info endpoint returns incorrect number of columns for linkfile

marikaris:
how to test:
- create a link
- see that the number of columns in the preview is the same as the number of variables selected

todo:
- [ ] added tests


timcadman (commented):
Works awesome, looks awesome!

Issue #124, opened 2021-04-09
404 message when saving workspace


timcadman:
```
datashield.workspace_save(conns, "bmi_poc_sec_2")
Error: Client error: (404) Not Found
```

timcadman:
No I've worked this one out - it was an error triggered by their not being sufficient disk space to save the workspace. It would be helpful if it could state this in the error (though not sure if you have control over this).

timcadman:
Updated issue to make clearer. As far as I know this is still an issue.

clemens-tolboom:
We have to check while saving a workspace we have enough storage

1. org.molgenis.armadillo.controller.DataController#saveUserWorkspace
1. https://github.com/molgenis/molgenis-service-armadillo/blob/e64e2994f245969889e689b8c245d5bee572c6c2/armadillo/src/main/java/org/molgenis/armadillo/command/impl/CommandsImpl.java#L197-L208
1. https://github.com/molgenis/molgenis-service-armadillo/blob/390d683a1f70043790e190e065e9545bf3243e86/armadillo/src/main/java/org/molgenis/armadillo/storage/LocalStorageService.java#L110-L126

timcadman:
Dick has checked and the largest workspace currently saved on BiB server is 1.3gb

StuartWheater:
You could try the algorithm:

1. Write the shadow workspace
2. Cause a "filesystem sync" to persistent storage
3. If (1) and (2) successful: Remove original workspace
4. If (3) successful: Rename shadow workspace to original workflow
5. Cause a "filesystem sync" to persistent storage
6. Repeat (4) and (5) until successful

This will require the running of a crash recovery process on machine restart, to ensure all shadow workspace have been renamed.

PR #738, opened 2024-05-16
fix: #124 throw error when workspace too big to store on device

marikaris:
Testing this is incredibly annoying. I tested it on the dev server and it kept dying, we really need to look into that. Maybe we should just trust that I've tested it, and just do the code review, if not, here is how I tested it:

How to test:
On the devserver, I put the build version of this PR, it's called `armadillo-workspace-4.x-SNAPSHOT.jar`
Check the version of armadillo on the devserver by going there in the browser and see what version is presented in the top menu, if it's not 4.3-SNAPSHOT (we have to investigate why my new build results in 4.3), then it's the incorrect version. To fix the version, ssh to demo server as root (see serverlist for IP)
``` bash
cd /usr/share/armadillo/application
service armadillo stop
rm armadillo.jar
ln -s armadillo-workspace-4.x-SNAPSHOT.jar armadillo.jar
service armadillo start
```
Check if server is back up. Open R studio. Create a script that creates a decently sized workspace:
``` R
library(MolgenisArmadillo)
library(DSMolgenisArmadillo)
library(dsBaseClient)
library(DSI)
demo_url <- "https://dev-armadillo.molgenis.org/"
demo_token <- armadillo.get_token(demo_url)

builder$append(
  server = "armadillo",
  url = demo_url,
  profile = "xenon",
  driver = "ArmadilloDriver",
  token = demo_token
)

logindata <- builder$build()

conns <- DSI::datashield.login(logins = logindata, assign = F)

datashield.assign.table(conns, "nonrep", "lifecycle/core/nonrep")

fit <- ds.glmSLMA(
  formula = "preg_no ~ 1+ height_mes_m + preg_fever + preg_gain_mes + ethn1_p + weight_f1 + smk_p + death_child_age",
  family = "gaussian",
  newobj = "test",
  dataName = "nonrep")


for (i in 1:1000) {
  ds.assign("test", paste0("test", i))
}


datashield.workspace_save(conns, "test_save")
```
Go to the data folder on the server to check if the workspace was saved properly:
```bash
cd /usr/share/armadillo/data
ls
```
cd to the `user-....` dir
``` bash
ls -la
```
You should see, something like:
`-rw-r--r-- 1 armadillo armadillo 296008668 May 23 12:50 armadillo:test_save.RData`

Now logout and reload the workspace:
``` R
datashield.logout(conns)

conns <- datashield.login(logindata, restore = "test_save")
ds.ls()
```
Should get many many objects.

Now open the browser, go to the "Insight" tab, select "Server metrics" and search for "free".  Check how much diskspace is left. Open your terminal again, where you still should be logged in to the server as root. Create a file that is the free diskspace - 500MB:
```
fallocate -l 5.2g /usr/share/armadillo/data/test.img
```
If you get a message with no space left on device, please immediately remove the file again and create a smaller one:
```
rm /usr/share/armadillo/data/test.img
```
Go to the browser, see the error message appearing that there's only a small amount of diskspace left. It should be less than 550MBs, but more than 0 bytes! If it's 0 bytes, quickly remove the file and retry because the profile might crash.

Open the logs on the server:
`tail -1000f /var/log/armadillo/armadillo.log`

Open R studio and save workspace again:
```
datashield.workspace_save(conns, "test_save")
```

This takes a while, but when it's done, check the logs, you should see something like:
```
024-05-23 14:33:00.816 [pool-2-thread-2|] INFO  o.m.armadillo.audit.AuditLogger - AuditEvent [timestamp=2024-05-23T14:33:00.814920662Z, principal=m.k.slofstra@umcg.nl, type=SAVE_USER_WORKSPACE_FAILURE, data={id=armadillo:test_save, sessionId=07B1ACE84E5AA51F922B4D2C8F251BF1, message=org.molgenis.r.exceptions.RExecutionException: org.molgenis.armadillo.exceptions.StorageException: Can't save workspace: workspace too big (282.1 MB), not enough space left on device. Try to make your workspace smaller and/or contact the administrator to increase diskspace., type=java.util.concurrent.CompletionException, roles=[ROLE_SU]}]
```

Now cd back to the data folder and verify the workspace is still the last saved version of it (time should be last time you saved and not your current try) and it should still be the same size.

We now need to catch this exception in the DSMolgenisArmadillo package. 

todo:
- [ ] updated docs in case of new feature (docs: https://github.com/molgenis/molgenis-r-datashield/pull/74)
- [x] added tests

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
@@ -1,0 +1,31 @@
+package org.molgenis.armadillo.storage;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import org.apache.commons.io.IOUtils;
+import org.molgenis.armadillo.exceptions.StorageException;
+
+public class ArmadilloWorkspace {
+  byte[] content;
+
+  public ArmadilloWorkspace(InputStream is) {
+    content = toByteArray(is);
+  }
+
+  private byte[] toByteArray(InputStream is) {
+    try {
+      return IOUtils.toByteArray(is);
+    } catch (IOException e) {
+      throw new StorageException("Unable to read workspace");
+    }
+  }
+
+  public long getSize() {
+    return content.length;
+  }
+
+  public ByteArrayInputStream createInputStream() {
+    return new ByteArrayInputStream(content);
+  }
+}
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
@@ -8,9 +8,7 @@ import static org.molgenis.armadillo.storage.StorageService.getHumanReadableByte
 import static org.springframework.http.MediaType.APPLICATION_JSON;
 import static org.springframework.http.MediaType.APPLICATION_OCTET_STREAM;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
+import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.Principal;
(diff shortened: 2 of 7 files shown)


armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
```diff
@@ -0,0 +1,31 @@
...
+public class ArmadilloWorkspace {
+  byte[] content;
+
+  public ArmadilloWorkspace(InputStream is) {
+    content = toByteArray(is);
+  }
+
+  private byte[] toByteArray(InputStream is) {
+    try {
+      return IOUtils.toByteArray(is);
```
mswertz, line 18: will this mean all storage will be in memory? Could we prevent that?

mswertz (approved):

PR #764, opened 2024-06-25
refactor: rock connection

marikaris:
This is a refactor of the rock connection class, using the spring `RestClient`, rather than the (deprecated) `RestTemplate`, as suggested by spring. This should prevent blocking calls and overall work more smoothly. 

Included apache in gradle, because default calls will go via jetty, which fails:
https://github.com/spring-projects/spring-boot/issues/38960

### DIFF
--- a/r/src/main/java/org/molgenis/r/rock/RockConnection.java
+++ b/r/src/main/java/org/molgenis/r/rock/RockConnection.java
@@ -17,6 +17,7 @@ import org.springframework.http.*;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.client.ResponseExtractor;
+import org.springframework.web.client.RestClient;
 import org.springframework.web.client.RestClientException;
 import org.springframework.web.client.RestTemplate;
 import org.springframework.web.util.UriComponentsBuilder;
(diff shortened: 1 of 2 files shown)


marikaris:
Codecov is failing, but it's a refactor of existing code. Can't find where it should have more tests.

r/src/main/java/org/molgenis/r/rock/RockConnection.java
```diff
@@ -163,15 +170,38 @@ private String getRSessionResourceUrl(String path) {
...
-    return new HttpHeaders() {
-      {
-        String auth = application.getUser() + ":" + application.getPassword();
-        byte[] encodedAuth = Base64.getEncoder().encode(auth.getBytes(StandardCharsets.UTF_8));
-        String authHeader = "Basic " + new String(encodedAuth);
-        add("Authorization", authHeader);
-      }
-    };
+  private String getAuthHeader() {
+    String auth = application.getUser() + ":" + application.getPassword();
```
mswertz, line 174: password?

mswertz: This is such that a request from rock can use the password to authenticate so we know if the proces (that armadillo started via rock) has access to the data (from that user)? 
Kinda scary, is there not a toke we could use instead? Or a way to at least encrypt this? Or is the internal communication also SSL? Because untrusted R code might now abuse it? Or to go even further, we could symply use a session number to reconnect because the password doesn't add much security?

mswertz (approved):

PR #765, opened 2024-07-01
chore: install release test deps in prepare.bash

marikaris:
how to test:

todo:
- [ ] added tests


marikaris:
Don't remember why I opened this PR. Will close until I need it

PR #766, opened 2024-07-01
feat: preview col names

timcadman:


marikaris:
Need to rethink this one. Does not support linkfiles yet.

marikaris (approved):

PR #773, opened 2024-08-13
fix: throw proper error message when workspace over 2GBs

### LINKED ISSUE #772: Armadillo crashes when saving workspace over 2GB (max int size in bytes)
### How to Reproduce
Save big workspace (larger than 2 GBs)

### Expected behavior
It succeeds, but I can't load it because that takes too much memory

### Observed behavior
Crash whenever my workspace will be above intsize in bytes. Want to put try catch around fix in #124 so saving won't be possible over 2GB (because load won't work anyways), but with a nice clear error. 

After fixing in this repo, also fix error message in R repo.

marikaris:
part of #772

how to test:
Run armadillo locally, then run in R:
``` R
library(MolgenisArmadillo)
library(DSMolgenisArmadillo)
library(dsBaseClient)
library(DSI)
demo_url <- "https://dev-armadillo.molgenis.org/"
demo_url <- "http://localhost:8080"
demo_token <- armadillo.get_token(demo_url)


builder$append(
  server = "armadillo",
  url = demo_url,
  profile = "xenon",
  driver = "ArmadilloDriver",
  token = demo_token
)


logindata <- builder$build()

conns <- DSI::datashield.login(logins = logindata, assign = T)

datashield.assign.table(conns, "nonrep", "lifecycle/core/nonrep")

fit <- ds.glmSLMA(
  formula = "preg_no ~ 1+ height_mes_m + preg_fever + preg_gain_mes + ethn1_p + weight_f1 + smk_p + death_child_age",
  family = "gaussian",
  newobj = "test",
  dataName = "nonrep")

start_time <- Sys.time()
for (i in 1:7000) {
  ds.assign("test", paste0("test", i))
  #silently_run(ds.assign("test", paste0("test", i)))
}
end_time <- Sys.time()
print(paste0("assigning took: ", end_time - start_time))

start_time <- Sys.time()
datashield.workspace_save(conns, "test_save4")
```
Look in the logs, see the error message tells the user that saving a workspace of more than 2GB is not possible.

Will have to fix the error getting through in R

todo:
- [x] added tests

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
@@ -7,7 +7,7 @@ import org.molgenis.armadillo.exceptions.StorageException;
 
 public class ArmadilloWorkspace {
   byte[] content;
-  public static final String workspaceTooBigError = "Unable to load workspace. Maximum supported workspace size is 2GB";
+  public static final String WORKSPACE_TOO_BIG_ERROR = "Unable to load workspace. Maximum supported workspace size is 2GB";
 
   public ArmadilloWorkspace(InputStream is) {
     content = toByteArray(is);
@@ -17,7 +17,7 @@ public class ArmadilloWorkspace {
     try {
       return IOUtils.toByteArray(is);
     } catch (OutOfMemoryError e) {
-      throw new StorageException(workspaceTooBigError);
+      throw new StorageException(WORKSPACE_TOO_BIG_ERROR);
     } catch (Exception e) {
       throw new StorageException("Unable to load workspace, because: " + e);
     }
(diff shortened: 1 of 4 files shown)


armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloWorkspace.java
```diff
@@ -1,13 +1,13 @@
...
 import java.io.InputStream;
 import org.apache.commons.io.IOUtils;
 import org.molgenis.armadillo.exceptions.StorageException;
 
 public class ArmadilloWorkspace {
   byte[] content;
+  public static final String WORKSPACE_TOO_BIG_ERROR = "Unable to load workspace. Maximum supported workspace size is 2GB";
 
   public ArmadilloWorkspace(InputStream is) {
     content = toByteArray(is);
```
mswertz, line 13: Can we instead do a while(read(10000 bytes){count++} and then not having to load all in memory but still count?

mswertz (approved):
Probably less memory extensive solution would be to retrieve the workspace twice via RServerConnection (ie Rock) in the CommandsImpl

image

But for now we do this.

PR #777, opened 2024-09-09
feat: expose actuator/prometheus endpoint

marikaris:
how to test:
- Start up armadillo
- Run this commandline: 
`curl 'http://localhost:8080/actuator/prometheus' -i -X GET`
No stacktraces will show up in the armadillo logs and the version of the R serve profiles are visible in the commandline output:
```
# HELP rserve_processes_current Current number of RServe processes on the R environment
# TYPE rserve_processes_current gauge
rserve_processes_current{environment="default",} 2.0
rserve_processes_current{environment="xenon",} 2.0
```

For code review:
- Check if this is secure enough. Removed pre authorise on class `ProfileService`, but put it back on all the methods of that class, so that the `.getAll` class can be called with `runAsSystem` in the `doWithConnection` method of `RProcessEndpoint`. I think that we don't leak anything that is not supposed to be leaked, but better to check twice. I'm for instance a bit worried about leakin the `seed`, but I think that that doesn't happen. 

todo:
- [ ] added tests

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/info/RProcessEndpoint.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/info/RProcessEndpoint.java
@@ -15,6 +15,8 @@ import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.stereotype.Component;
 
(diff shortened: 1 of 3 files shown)


mswertz (approved):
Please take a look to the runAsSystem code if we could get this to work using the class level auth instead of function level auth. Don't spend more then 30mins, otherwise this is ok.

marikaris:
Can't get this to deploy on dev server. (Same for master)

Issue #774, opened 2024-09-06
When user is not an admin, unclear error appears in UI


marikaris:
### How to Reproduce
Go to ui as non-admin

### Expected behavior
Error message telling me I'm not an admin, so I cannot use the UI

### Observed behavior
ApiError

PR #802, opened 2024-10-14
fix: #774 when unauthorised, don't show tabs and give clear error mes…

marikaris:
…sage

how to test:

todo:
- [ ] added tests


mswertz (approved):
code okay

PR #803, opened 2024-10-15
feat: form validation link file

marikaris:
how to test:
- go to [UI](https://preview-armadillo-pr-803.dev.molgenis.org/)
- create link file 
- fill in nothing
- click on save
- see form validation

todo:
- [ ] added tests


erikzwart (changes requested):
Functional test on preview, created a Project cohort and folder test. Uploaded nonrep.parquet. The buton 'Create view' does not work:
image
After selecting 'Create view':
image

Also tried the button 'Select table to link from ..', here I can select the source project, folder and table but after that I see the same empty screen as above.

I'm not able to test the form validation because the form is not shown.

timcadman (approved):

PR #813, opened 2024-10-25
feat: reimplement rserve route

marikaris:
how to test:

todo:
- [ ] added tests => e2e test should work on both Rserve and Rock, might need additional profile


marikaris:
Sonar is failing, but coverage has been increased a lot. The added in code, is mostly old code that we're reintroducing. This code doesn't have many tests. Mostly because the complexity of tests that would have to be written, writing those tests would take up more time than having them would ever grant us.

mswertz (commented):
Code looks good. Would like a quick chat for final acceptance.

Is it possible to reuse exact tests between Rock and Rserve using a parameter to run all test twice using either or? Maybe even a totally outside parameter via gradle. Or is this maybe we what I see.

marikaris:
> Code looks good. Would like a quick chat for final acceptance.
> 
> Is it possible to reuse exact tests between Rock and Rserve using a parameter to run all test twice using either or? Maybe even a totally outside parameter via gradle. Or is this maybe we what I see.

I think the rock code lacks in the amount of tests it has. I have reintroduced the tests we had for rserve and added a couple new tests. I have tested both paths using our releasetest, using both profiles (rserve/rock) and that test succeeds both ways.

mswertz:
we discussed: use e2e test and then run on both.
Will probably require a specific profile in the preview server.

PR #816, opened 2024-11-14
docs: add new docs structure

marikaris:
# What have I done?
- Removed the old docs folder. We need to look at our old docs on the master and see what we can copy paste in this new structure from there.
- Initialised mkdocs
- Added in agreed upon structure. Each file in the menu has the key words that should be in the article in them. 
- Added css that somewhat matches molgenis theme

# Test running docs
Mkdocs allows for seeing what the docs will look like locally, if you prefer to do so. You do need to have python and pip installed, as well as mkdocs: 
```
pip install mkdocs
pip install mkdocs-material
pip install pymdown-extensions
```
After that, simply run this in the armadillo repository:
```
mkdocs serve
```
The docs will then run on:
http://localhost:8000/

## MkDocs Material Theme

See [getting started](https://squidfunk.github.io/mkdocs-material/getting-started/) and [reference](https://squidfunk.github.io/mkdocs-material/reference/) formatting options for the material theme for MKDocs

# To do 
**(please put your initials to the point if you're working on it)**

- [x] Index (MkS)
- [x] Quick start (MkS)
- [ ] About us (There's something there, but it needs more)
- [x] Basic concepts (MkS)
- [x] Basic usage (ez)
- [ ] Example of usage (We could skip this for now?)
- [ ] Advanced usage (We could skip this for now?)
- [x] Installation, setup and managment (ez)
- [x] Developer guides (MkS)
- [x] License (ez)
- [x] FAQ (ez)
- [ ] Troubleshooting
- [x] Contact (ez)
- [ ] Go through old md files scattered throughout repo and move things into docs + link to docs
- [ ] Build docs automatically
- [ ] Link to built docs in UI
- [ ] Add cite us message to readme + docs
- [ ] Rewrite readme to include very quick start


timcadman:
A few style notes: 
- We should be consistent whether we refer to Armadillo or armadillo I would vote capitalised?
- We should be consistent between clientside, client side or client-side (same for server-side). I would suggest 'client-side & server-side'
- We should refer to our packages using exactly the same terms as the paper, ie 
  - molgenis-service-armadillo = armadilloService
  - molgenis-r-armadillo = MolgenisArmadillo
  - molgenis-r-datashield = DSMolgenisArmadillo

mswertz (approved):

Issue #751, opened 2024-06-18
As Armadillo datamanager I want to manage saved workspaces in the UI


marikaris:
Motivation: 
* Workspaces are now a bit hidden away in the data folder of armadillo and hard for the datamanagers to access 

Demo:
* Datamanager can see all workspaces of their server in the userinterface
* Datamanager can see how big all these workspaces are
* Datamanager can see who the owner of the workspace is
* Datamanager can delete workspace

Acceptance criteria:
* Expose workspaces via API for admins
* Show workspaces in UI with: user and size
* Add button for deleting workspace (with confirmation dialog)

PR #820, opened 2024-12-10
Feat: #751 show all workspaces to admin

marikaris:
New PR because old one got tangled up with the docs PR for some reason. 

how to test:

todo:
- [ ] added tests

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
@@ -15,6 +15,7 @@ import java.security.Principal;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.commons.io.FilenameUtils;
 import org.molgenis.armadillo.exceptions.*;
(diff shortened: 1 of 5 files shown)


marikaris:
Ran into this bug:
https://github.com/molgenis/molgenis-service-armadillo/issues/821

The delete workspace endpoint has the same problem. I think this bug needs to be fixed in this PR as well.

marikaris:
I'm suspecting the quality gate above is the right one, not the one in the "checks" part, as that one shows files that were not touched.

armadillo/src/main/java/org/molgenis/armadillo/audit/AuthenticationAuditListener.java
```diff
@@ -29,7 +28,8 @@ public void onApplicationEvent(@NotNull AbstractAuthenticationEvent event) {
   private void onAuthenticationSuccessEvent(AuthenticationSuccessEvent event) {
     publish(
         new AuditEvent(
-            getUser(event.getAuthentication().getPrincipal()),
+            UserInformationRetriever.getUserIdentifierFromPrincipal(
```
mswertz, line 31: waarom heb je deze nieuwe functie niet ook gewoon getUser genoemd? Dat had flink wat code changes verborgen ;-)

armadillo/src/main/java/org/molgenis/armadillo/storage/ArmadilloStorageService.java
```diff
@@ -202,21 +221,38 @@ public List<Workspace> listWorkspaces(Principal principal) {
...
-      throw new IllegalArgumentException(
-          "Cannot create valid S3 bucket for username " + principal.getName());
-    }
-    return bucketName;
+  private static String getOldUserBucketName(Principal principal) {
+    return USER_PREFIX + principal.getName();
+  }
+
+  static String getUserBucketName(Principal principal) {
+    String userIdentifier = getUserIdentifierFromPrincipal(principal).replace("@", "__at__");
```
mswertz, line 254: you tested against weird email adresses?

mswertz (commented):
code looks good, jus small comments.
Will test now.

N.B. I understood file system doesn't care about special characters like umlauts. But you might need to address these? [- _ + ! # $ % & ' * / = ? ^ { | } ~] which as I understand are technically legal or at least confusing as they have special meanings.?

mswertz (approved):

PR #846, opened 2025-04-24
feat: csv/tsv import

marikaris:
how to test:
- Open armadillo UI
- Go to project
- Upload CSV
- See csv file uploaded as parquet

Todo:
- [x] Read csv file
- [x] Determine types 
- [x] Create parquet file
- [x] Write to parquet file
- [x] Tests
- [x] int32
- [x] boolean

Notes: 
In R we convert a dataframe with all the types already set to a parquet file, resulting in the right datatypes. We first read the csv into a dataframe, then set the types where needed and then it's converted. 
Solution: Determine the types from the first 100 rows by default. User can define to use more or less rows to determine types from.


armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -116,14 +117,58 @@ public void uploadObject(
...
+  @Operation(summary = "Upload a csv file to a project")
+  @ApiResponses(
+      value = {
+        @ApiResponse(responseCode = "204", description = "Object uploaded successfully"),
+        @ApiResponse(responseCode = "404", description = "Unknown project"),
+        @ApiResponse(responseCode = "409", description = "Object already exists"),
+        @ApiResponse(responseCode = "401", description = "Unauthorized")
+      })
+  @PostMapping(
+      value = "/projects/{project}/objects/csv",
```
mswertz, line 129: we will not have problems with other meanings of this path? often the path is GET {project}/objects/{id} and thus also posts? I conflicts are an issue the  {project}/csv/{id} might bet better. Please advice?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -116,14 +117,58 @@ public void uploadObject(
...
+  public void uploadCharacterSeparatedFile(
+      Principal principal,
+      @PathVariable String project,
+      @RequestParam @NotEmpty String object,
+      @RequestParam int numberOfRowsToDetermineTypeBy,
+      @Valid @RequestParam MultipartFile file) {
+    auditor.audit(
+        () -> {
+          try {
+            addParquetObject(project, object, file, numberOfRowsToDetermineTypeBy);
```
mswertz, line 141: I had expected that you would have a 'csvToParquet' somewhere...

mswertz: this function is 'convertToCsvAndAddAsParquet? 
I will see :-)

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+        && reader.getLinesRead() < this.numberOfRowsToDetermineTypeBy) {
+      int i = 0;
+      for (String value : line) {
+        // determine type
+        // if value is NA, it's not a string by definition
+        if (!value.isEmpty()) {
+          String type = getTypeOfCell(value);
+          // only set to double if is not string yet, because if one value is not double, the cell
+          // cannot be double and must be string
+          if (Array.get(types, i) != STRING && Objects.equals(type, DOUBLE)
```
mswertz, line 121: we only have double and string in parquet? no int, no categorical?

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+            + "\"name\": \"armadillo\"," // Not used in Parquet, can put anything
+            + "\"fields\": ";
+    ArrayList<String> fields = new ArrayList<>();
+    AtomicInteger emptyCols = new AtomicInteger(1);
+    final int[] colNumber = {0};
+    Arrays.stream(header)
+        .forEach(
+            (headerValue) -> {
+              // replace BOM if present
+              headerValue = headerValue.replace("\uFEFF", "");
```
mswertz, line 145: well spotted

mswertz: BOM is often an issue

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+  static String STRING = "string";
+  static String DOUBLE = "double";
+  MultipartFile file;
+  char separator = ',';
+  Schema schema;
+  String[] header;
+  List<String> types;
+  int numberOfRowsToDetermineTypeBy = 100;
+
+  public CharacterSeparatedFile(MultipartFile file) throws IOException, CsvValidationException {
```
mswertz, line 38: below, it was unclear what methods are supposed to be private

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+
+  public void setHeader(String[] header) {
+    this.header = header;
+  }
+
+  public void setHeaderColname(String colname, int index) {
+    this.header[index] = colname;
+  }
+
+  static char getSeparatorFromHeader(String[] header, MultipartFile file)
```
mswertz, line 70: private?

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  private CSVReader getReader() throws IOException {
+    CSVParser parser = new CSVParserBuilder().withSeparator(separator).build();
+    return new CSVReaderBuilder(new InputStreamReader(file.getInputStream()))
+        .withCSVParser(parser)
+        .build();
+  }
+
+  static String getTypeOfCell(String cell) {
```
mswertz, line 99: private?

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+    try {
+      Double.parseDouble(cell);
+      return CharacterSeparatedFile.DOUBLE;
+    } catch (NumberFormatException e) {
+      return CharacterSeparatedFile.STRING;
+    }
+  }
+
+  List<String> getTypesFromData(CSVReader reader) throws IOException, CsvValidationException {
```
mswertz, line 108: private?

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+            Array.set(types, i, type);
+          }
+        }
+        i++;
+      }
+    }
+    return Arrays.stream(types).map((type) -> Objects.requireNonNullElse(type, STRING)).toList();
+  }
+
+  Schema createSchemaFromTypes(List<String> types, String[] header) {
```
mswertz, line 132: private?

armadillo/src/test/java/org/molgenis/armadillo/controller/StorageControllerTest.java
```diff
@@ -89,6 +90,108 @@ void uploadObject() throws Exception {
             mockSuAuditMap(Map.of(PROJECT, "lifecycle", OBJECT, "core/nonrep2.parquet"))));
   }
 
+  @Test
+  void uploadCsvObject() throws Exception {
```
mswertz, line 94: rename to testUploadCsvObject ? or don't we prefix all with 'test' in Armadillo?

armadillo/src/test/java/org/molgenis/armadillo/storage/CharacterSeparatedFileTest.java
```diff
@@ -0,0 +1,113 @@
...
+  void testConstructorWithCommaSeparator() throws IOException, CsvValidationException {
+    String csvData = "name,age\nJohn,30\nJane,25\n";
+    Mockito.when(mockFile.getInputStream())
+        .thenReturn(new ByteArrayInputStream(csvData.getBytes()));
+    Mockito.when(mockFile.getOriginalFilename()).thenReturn("test.csv");
+
+    CharacterSeparatedFile csf = new CharacterSeparatedFile(mockFile);
+
+    assertNotNull(csf);
+    assertEquals(',', csf.separator);
```
mswertz, line 39: probably should be getters and the properties private? Or are these properties immutible?

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
```diff
@@ -0,0 +1,207 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.parquet.avro.AvroParquetWriter;
+import org.apache.parquet.hadoop.ParquetWriter;
+import org.apache.parquet.hadoop.metadata.CompressionCodecName;
+import org.apache.parquet.io.LocalOutputFile;
+import org.molgenis.armadillo.exceptions.FileProcessingException;
+import org.springframework.web.multipart.MultipartFile;
+
+public class CharacterSeparatedFile {
+  static String STRING = "string";
```
mswertz, line 29: private properties? I understand you do this for testability? but then you could have side effects. Or is that okay? It is a bit unconventional ... ;-)

mswertz (changes requested):
maybe you can explain about the friendly, non-private methods and properties?

Will this become a problem when set from outside?

armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
marikaris: I based the endpoint on the objects endpoint (which uploads any file, not just csv), but maybe `/projects/{project}/csv` would be better. I didn't add in the id in the URL for consistency (because the `/objects` endpoint doens't have it in there).

armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
marikaris: Yes, I have looked at parquet files that have categoricals/ints/doubles in them and found out only two datatypes of the parquet files were used.

PR #863, opened 2025-06-05
fix: also take typing into account when writing parquet file

marikaris:
how to test:

todo:
- [ ] added tests


erikzwart (changes requested):
On the preview I try to upload the test_csv.csv. The UI throws a api error, however when I retry to upload test_csv.csv I get this api error: 
image

So the file seems to be uploaded but not correctly converted to parquet?
Test file used: 
[test_csv.csv](https://github.com/user-attachments/files/20613317/test_csv.csv)

PR #872, opened 2025-07-04
feat: auto update profile

timcadman:
# What has changed
- Implemented a scheduler to check whether currently running profiles have been updated
- extended the profileConfig class with two new fields controlled via UI: (i) auto-update (boolean), (ii) frequency of autoupdate
- Implemented functionality in front end

# How to test
- [ ] Code review
- [ ] UI review
- [ ] Functional tests (below)

### Test that selecting autoupdate in front end triggers back-end check
- Login to UI
- For an existing profile, click 'edit', check 'auto-update', and in the drop-down box select 'daily' and a time around 1 minute from now
- Check that the Armadillo log displays `Scheduled auto-update for profile [xxx] with cron [xxx]
- Check that at the time you scheduled the autoupdate, the log reads `INFO  o.m.a.profile.ProfileScheduler - No image update for 'xxx', skipping restart`

## How to test:
- Create an account in [docker hub](https://hub.docker.com/) if you don't already have one
- Create a folder to hold test docker images and navigate to that folder

Enter your username and password

To be 100% sure of what is going on, we need to isolate the docker environment used to create the images from the localhost environment. We can use docker-in-docker for this. In terminal run:

```
docker run --privileged -v /Users/tcadman/repos/testing/docker -d --name docker-build-env docker:dind
```
Enter the container:
```
docker exec -it docker-build-env /bin/sh
mkdir /scripts
cd scripts
```
Now log in to docker
```
docker login
```
Open a browser with the address displayed in the terminal and enter the code.

Return to the terminal and run the following:
```
touch build-and-push.sh
chmod +x build-and-push.sh
```

Edit the file
```
vi build-and-push.sh
```

Now add this script, which will can be used to create a very minimal docker image for testing (note you need to change 'timmyjc' to your docker usename)
```
#!/bin/sh

VERSION="$1"

if [ -z "$VERSION" ]; then
  echo "Usage: $0 <version>"
  exit 1
fi

IMAGE_NAME="timmyjc/mytest"

# Create Dockerfile
cat <<EOF > Dockerfile
FROM alpine
LABEL org.opencontainers.image.version="$VERSION"
CMD ["sh", "-c", "while true; do echo Hello from v$VERSION; sleep 30; done"]
EOF

# Build image with version tag
docker build -t $IMAGE_NAME:v$VERSION .

# Tag it as latest
docker tag $IMAGE_NAME:v$VERSION $IMAGE_NAME:latest

# Push both tags
docker push $IMAGE_NAME:v$VERSION
docker push $IMAGE_NAME:latest
```
Save and exit
```
:wq
```
```
Check there are no images
```
docker images
```

Now create a new image using:
```
./build-and-push.sh 0.0.1
```

Now go to the UI, and add the image that you have just created and start the profile. Check that the version number is displaying 0.0.1.

Once the image has started, go back to DinD and create a new image with a new version number:
```
 ./build-and-push.sh 0.0.2
```

Return to the UI, edit the profile, select 'auto-update', and schedule for 1 minute from now. At the time you scheduled the update, check in the logs that the following is displayed:

```
Latest imageId is xxx
Image ID for profile 'xxx' changed from 'xxx' to 'xxx' 
Image updated for 'xxx', restarting...
Image ID for profile 'test' changed from 'xxx' to 'xxx'
Removed image ID 'xxx' from local Docker cache
```
Return to the UI, refresh the page, and check that version 0.0.2 is now displaying.

todo:
- [ ] update docs with details of new feature

### CODE REVIEW

File: armadillo/src/main/java/org/molgenis/armadillo/profile/ProfileScheduler.java [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+
+    if ("weekly".equalsIgnoreCase(frequency)) {
+      int dayOfWeek = dayToCronNumber(day != null ? day : "Sunday");
+      return String.format("0 %s %s * * %d", minute, hour, dayOfWeek);
+    } else {
+      return String.format("0 %s %s * * *", minute, hour); // daily or fallback
+    }
+  }
+
+  private int dayToCronNumber(String day) {
```
marikaris, line 85: Function name should be verb, maybe `convertDayToCronNumber` or `getCronNumberFromDay`?

PR #876, opened 2025-07-28
feat: all docker images endpoint

marikaris:
how to test:
- Install armadillo locally
- Make sure docker is running with some images pulled in there
- Go to new swagger endpoint: http://localhost:8080/swagger-ui/index.html#/insight/getDockerImages
- Click on Execute and see all images there (including non-running)

todo:
- [x] updated docs in case of new feature (Swagger should be sufficient for API endpoint?)
- [x] added tests


mswertz (approved):
works on my machine too. code looks okay. security is in place.

PR #897, opened 2025-09-09
feat: show profile status

timcadman:
## Issue:
Currently when pulling a new image (especially a large one) no information is returned to the user, which can leave them unsure what is actually happening.

## What's changed
Install status is now exposed as number of layers installed and total number of layers (it is hard/impossible to get anything more accurate from docker). This is implemented as a progress bar in the front end, with some smoothing trickery applied  (without this it spends a long time stuck on certain values when a large layer is being installed).

## how to test:
- Add a new profile (make sure it pulls an image that is not currently downloaded)
- Check the profile bar displays whilst the image is pulling and disappears when it has finished
- Start a profile which already exists and check that nothing strange happens
- Code review


timcadman:
- [ ] Could you show an actual progressbar? You can use this one: https://getbootstrap.com/docs/5.3/components/progress/#bar-sizing, I’m already using it in the Actuator.vue for the diskspace
- [ ] Maybe instead of “Installing layer” I’d say “Pulling docker image”, or something else, “Installing profile” maybe.
- [ ] I just tested, adding 1 profile, starting it, it will start installing, then I add another one and start installing that one, I lose progress on the first one. Both do eventually complete and start :smile:
- [ ] Maybe show which profile it’s installing as well (otherwise it can be confusing when 2 things are installing at once)
- [ ] Maybe instead of showing “Starting” Below the start/stop button, show “Installing” when it’s still installing

ui/src/components/ProfileStatusMessage.vue
```diff
@@ -0,0 +1,64 @@
...
+  if (!s) return 0;
+  const total = s.totalLayers ?? 0;
+  const completed = s.completedLayers ?? 0;
+  return total > 0 ? Math.round((completed / total) * 100) : 0;
+});
+
+const smoothedPerc = ref(0);
+const displayedPerc = computed(() => Math.round(smoothedPerc.value));
+
+function smoothServerPercentage() {
```
marikaris: I think this one deserves a unittest

marikaris (changes requested):
I really like this functionality. Just missing tests for one function in the frontend

PR #963, opened 2026-02-03
feat: add bearer token to resource serving

mswertz:
To do:
- [x] Make endpoint available when using internally generated token
- [x] Add tests
- [x] Test if resources still work 
- [x] Run release test
- [x] Write docs

### DIFF
--- a/scripts/release/testthat/tests/test-10-resources.R
+++ b/scripts/release/testthat/tests/test-10-resources.R
@@ -1,0 +1,161 @@
+# Setup
+test_name <- "resources"
+resource_path <- "ewas/GSE66351_1"
+
+# Helper: reconnect DataSHIELD session (failed operations can crash the R session)
+reconnect <- function() {
+  tryCatch(datashield.logout(release_env$conns), error = function(e) NULL)
+  logindata <- suppressWarnings(create_dsi_builder())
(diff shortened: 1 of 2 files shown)


armadillo/src/main/java/org/molgenis/armadillo/controller/StorageController.java
```diff
@@ -387,18 +386,72 @@ public void deleteObject(
...
+  ResponseEntity<InputStreamResource> downloadResourceWithToken(
+      JwtAuthenticationToken token, String project, String object, Map<String, Object> data) {
+    Map<String, Object> claims = token.getTokenAttributes();
+    String errorMsg = "Token must be issued by armadillo application with correct permissions";
+    if (!claims.get("iss").equals(INTERNAL_ISSUER)) {
+      auditFailure(errorMsg, data, token);
+      throw new ResponseStatusException(HttpStatus.FORBIDDEN, errorMsg);
+    }
+    if (claims.get("resource_project").equals(project)) {
+      String resourceObj = object.split("\\.")[0].toLowerCase();
```
timcadman: Same issue above, will fail if resource has a `.` in the name. Also consider abstracting to common helper method

PR #973, opened 2026-02-14
Test/setup packages

### LINKED ISSUE #650: release test not successfully whitelisting packages
- [ ] ? relates #651 
- [ ] ? revert CI fix adding dsMediation https://github.com/molgenis/molgenis-service-armadillo/blob/master/docker/ci/application.yml#L34

---

Expected that this would set packages to be whitelisted:

```
profile_defaults = data.frame(
  name = c("xenon", "rock"),
  container = c("datashield/rock-dolomite-xenon:latest", "datashield/rock-base:latest"),
  port = c("", ""),
  # Multiple packages can be concatenated using ,, then using stri_split_fixed() to break them up again
  # Not adding dsBase since that is always(?) required
  whitelist = c("resourcer,dsMediation,dsMTLBase", ""),
  blacklist = c("", "")
)
```
However CI failed because packages (e.g. dsMediation) not present. Workaround is to add package to docker/ci/application.yml

### LINKED ISSUE #958: Release test tries to run tests for packages that don't exist in current profile

timcadman:
## What's changed
- Automatically starts profile if stopped
- Automatically whitelists packages if they're available in the profile
- Skips test automatically if the package is not available for that test

## How to test
Run release for following cases:
- [x] No profile specified in .env file, should test on default and skip tests not in default profile.
- [x] Default profile specified in .env file, which is started and has the packages whitelisted. Should test on default and skip tests not in default profile.
- [x] Default profile specified in .env file, not started and no packages whitelisted. Should start profile, whitelist packages available in this profile and skip tests for packages not available in this profile.
- [x] Default & Donkey profiles both specified in .env file, both started and packages whitelisted. Should first run tests on default, skipping for packages not available, then run on Donkey for all packages.
- [x] Default & Donkey profiles both specified in .env file, not started and no packages whitelisted. As above, but first will start and whitelist packages.


marikaris (dismissed):

marikaris (changes requested):
CI is failing because of something with xenon?

timcadman:
> CI is failing because of something with xenon?

Fixed. CI now just failing on critical CVE which is addressed in a different PR

marikaris (approved):

PR #985, opened 2026-03-02
feat(flwr) #001: containers

timcadman:
## Background
Currently implemented container types are Vanilla and DataSHIELD. To run Flower on Armadillo we need (i) Flower-specific container classes and (ii) updated methods to pass parameters when starting containers.

## New container types
- `FlowerSupernodeContainerConfig` — config for Flower supernode containers, with `caCertPath` and `authPrivateKeyPath` fields for mounting the TLS root certificate and supernode auth key
- `FlowerSuperexecContainerConfig` — config for Flower superexec/clientapp containers, with a `fabWhitelistPath` field for mounting the FAB-hash whitelist (the app-trust mechanism enforced by the SuperExec plugin)
- Both implement `FlowerContainer`, a marker interface used in `instanceof` checks and holding the `flower-network` constant

## DockerService changes
- Port binding is skipped for Flower containers, which communicate over the Docker network, not host ports
- Superexec containers get an `ARMADILLO_CONTAINER_NAME` env var (for callbacks to Armadillo when pushing data) and, when `flower.armadillo-url` is configured, an `ARMADILLO_URL` env var (the callback address itself)
- `configureBindMounts` mounts files read-only into Flower containers:
  - Supernode: `ca.crt` (root CA for TLS to the SuperLink) and `credentials` (supernode auth private key)
  - Superexec: the FAB-hash whitelist YAML
- Cert/key files are validated at start time: must exist and be non-empty, failing closed. The FAB whitelist file is the one exception — empty is allowed, since "nothing approved yet" is a legitimate state
- `dockerArgs` are passed as container CMD arguments
- Superexec containers always run with entrypoint `armadillo-flwr-superexec` (https://github.com/molgenis/molgenis-flwr-armadillo/pull/11), so only an image containing the FAB-whitelist-enforcing SuperExec can run as a clientapp; the stock `flwr/superexec` image fails to start.

## ContainerService changes
- On save, placeholder files are created at `data/system/flower/` for the supernode certificate/credentials and the superexec FAB whitelist, if they don't already exist
- Admin must replace empty cert/credential placeholders with real values before starting the container.

## Interface changes
- `ContainerConfig`: `host` and `port` made nullable (Flower containers don't expose host ports)
                                                                                                                                              
### how to test:
- [x] Code review
- [ ] Check CI green
- [ ] Start Armadillo on localhost and follow steps below to register and start supernode and superexec.

```bash
# Add to data/application.yml:
#   flower:
#     armadillo-url: <URL at which the containers can reach Armadillo; needed to push data into the container>

# 1. Navigate to python package and build the clientapp image from `feat/verified-superexec` branch
docker build -f docker/superexec-armadillo.Dockerfile -t superexec-armadillo:test .

# 2. Navigte back to molgenis-service-armadillo and start from an empty Flower directory
rm -rf data/system/flower/

# 3. Create the container definitions
jq -n '{type:"flower-supernode", name:"flower-supernode-1", image:"flwr/supernode:1.32.1",
 dockerArgs:["--insecure","--superlink","superlink.molgenis.net:9092"]}' > /tmp/supernode.json
jq -n '{type:"flower-superexec", name:"flower-clientapp-1", image:"timmyjc/superexec-armadillo:test",
 dockerArgs:["--insecure","--appio-api-address","flower-supernode-1:9094"]}' > /tmp/clientapp.json

# 4. Register both (creates empty certificate, key and whitelist files)
curl -u admin:admin -X PUT http://localhost:8080/containers -H "Content-Type: application/json" -d @/tmp/supernode.json
curl -u admin:admin -X PUT http://localhost:8080/containers -H "Content-Type: application/json" -d @/tmp/clientapp.json
# expect: HTTP 204 for both

# 5. Fill the certificate and key files (the whitelist may stay empty: no apps approved yet)
echo "dummy-ca" > data/system/flower/ca.crt
echo "dummy-key" > data/system/flower/credentials

# 6. Start both
curl -u admin:admin -X POST http://localhost:8080/containers/flower-supernode-1/start
curl -u admin:admin -X POST http://localhost:8080/containers/flower-clientapp-1/start
# expect: HTTP 204 for both

# 7. Check they are running, and that the clientapp uses the whitelist-enforcing entrypoint
docker ps --filter name=flower-supernode-1 --filter name=flower-clientapp-1
docker inspect flower-clientapp-1 --format '{{json .Config.Entrypoint}} {{json .Config.Cmd}}'
# expect: ["armadillo-flwr-superexec"] ["--fab-whitelist","/app/fab-whitelist.yaml","--insecure","--appio-api-address","flower-supernode-1:9094"]

# 8. Check the stock image is refused
jq '.name = "flower-clientapp-stock" | .image = "flwr/superexec:1.32.1"' /tmp/clientapp.json > /tmp/clientapp-stock.json
curl -u admin:admin -X PUT http://localhost:8080/containers -H "Content-Type: application/json" -d @/tmp/clientapp-stock.json
curl -u admin:admin -X POST http://localhost:8080/containers/flower-clientapp-stock/start
# expect: an error; the container does not start ("armadillo-flwr-superexec": executable file not found)

# 9. Clean up
docker rm -f flower-supernode-1 flower-clientapp-1 flower-clientapp-stock
docker network rm flower-network
rm -rf data/system/flower/
rm /tmp/supernode.json /tmp/clientapp.json /tmp/clientapp-stock.json
```


armadillo/src/main/java/org/molgenis/armadillo/container/DockerService.java
```diff
@@ -250,27 +269,138 @@ boolean hasImageIdChanged(String containerName, String previousImageId, String c
...
+      hostConfig.withNetworkMode(FLOWER_NETWORK_NAME);
+    }
+  }
+
+  private void configureEnv(CreateContainerCmd cmd, ContainerConfig config) {
+    if (config instanceof FlowerSuperexecContainerConfig) {
+      if (flowerArmadilloUrl == null || flowerArmadilloUrl.isEmpty()) {
+        throw new IllegalStateException(
+            "flower.armadillo-url is not configured — required so Flower clientapp containers"
+                + " can reach Armadillo, please set it in application.yml");
```
marikaris, line 319: Do we want to add this to the sytem UI + api?

armadillo/src/main/java/org/molgenis/armadillo/container/FlowerSuperexecContainerConfig.java
```diff
@@ -0,0 +1,145 @@
...
+  @Override
+  public abstract String getImage();
+
+  @Override
+  @Nullable
+  public abstract String getHost();
+
+  @Override
+  @Nullable
+  public abstract Integer getPort();
```
marikaris: I thought the port is set to null in flower containers? Is this so that it will fit the ContainerConfig?

armadillo/src/main/java/org/molgenis/armadillo/metadata/ContainerService.java
```diff
@@ -73,12 +76,39 @@ public void upsert(ContainerConfig containerConfig) {
...
+
+  private void createFileIfNotExists(String pathStr) {
+    if (pathStr == null) return;
+    Path path = Path.of(pathStr);
+    if (Files.exists(path)) return;
+    try {
+      Files.createDirectories(path.getParent());
+      Files.createFile(path);
+    } catch (IOException e) {
+      throw new IllegalStateException("Failed to create placeholder file: " + path, e);
```
marikaris, line 108: If this function is for creating placeholder files, maybe put it in the method name?

armadillo/src/main/java/org/molgenis/armadillo/container/FlowerSuperExecContainerConfig.java
marikaris, line 1: This one is very similar to FlowerSupernodeContainer config. I would suggest creating an abstract class that is extended in both the FlowerSupernodeContainer and FlowerSuperexecContainer. I would also prefer the names to be: FlowerSuperNodeContainer and FlowerSuperExecContainer. Asked claude to do it, haven't tested, but it might help:
```
package org.molgenis.armadillo.container;

import jakarta.annotation.Nullable;
import java.util.List;
import java.util.Map;

public abstract class AbstractFlowerContainerConfig
    implements ContainerConfig, OpenContainer, FlowerContainer {

  @Override
  public abstract String getName();

  @Override
  public abstract String getImage();

  @Override
  @Nullable
  public abstract String getHost();

  @Override
  @Nullable
  public abstract Integer getPort();

  @Override
  @Nullable
  public abstract Long getImageSize();

  @Override
  @Nullable
  public abstract String getInstallDate();

  @Override
  @Nullable
  public abstract String getLastImageId();

  @Override
  @Nullable
  public abstract List<String> getDockerArgs();

  @Override
  @Nullable
  public abstract Map<String, Object> getDockerOptions();

  @Override
  @Nullable
  public abstract String getVersionId();

  @Override
  @Nullable
  public abstract String getCreationDate();

  // Self-typed generic builder base so setters can return the concrete builder type.
  public abstract static class Builder<B extends Builder<B>> {
    public abstract B name(String name);

    public abstract B image(String image);

    public abstract B host(@Nullable String host);

    public abstract B port(@Nullable Integer port);

    public abstract B imageSize(@Nullable Long imageSize);

    public abstract B installDate(@Nullable String installDate);

    public abstract B lastImageId(@Nullable String lastImageId);

    public abstract B dockerArgs(@Nullable List<String> dockerArgs);

    public abstract B dockerOptions(@Nullable Map<String, Object> dockerOptions);

    public abstract B versionId(@Nullable String versionId);

    public abstract B creationDate(@Nullable String creationDate);
  }
}
```

```
package org.molgenis.armadillo.container;

import com.fasterxml.jackson.annotation.*;
import com.google.auto.value.AutoValue;
import jakarta.annotation.Nullable;
import java.util.List;
import java.util.Map;

@AutoValue
@JsonTypeName("flower-supernode")
@SuppressWarnings({"java:S1161", "java:S3038"})
public abstract class FlowerSupernodeContainerConfig extends AbstractFlowerContainerConfig {

  @Nullable
  public abstract String getCaCertPath();

  @Nullable
  public abstract String getAuthPrivateKeyPath();

  @Override
  @JsonIgnore
  public String getType() {
    return "flower-supernode";
  }

  @JsonCreator
  public static FlowerSupernodeContainerConfig create(
      @JsonProperty("name") String name,
      @JsonProperty("image") String image,
      @JsonProperty("host") @Nullable String host,
      @JsonProperty("port") @Nullable Integer port,
      @JsonProperty("imageSize") @Nullable Long imageSize,
      @JsonProperty("installDate") @Nullable String installDate,
      @JsonProperty("lastImageId") @Nullable String lastImageId,
      @JsonProperty("dockerArgs") @Nullable List<String> dockerArgs,
      @JsonProperty("dockerOptions") @Nullable Map<String, Object> dockerOptions,
      @JsonProperty("versionId") @Nullable String versionId,
      @JsonProperty("creationDate") @Nullable String creationDate,
      @JsonProperty("caCertPath") @Nullable String caCertPath,
      @JsonProperty("authPrivateKeyPath") @Nullable String authPrivateKeyPath) {

    return builder()
        .name(name)
        .image(image)
        .host(host)
        .port(port)
        .imageSize(imageSize)
        .installDate(installDate)
        .lastImageId(lastImageId)
        .dockerArgs(dockerArgs)
        .dockerOptions(dockerOptions)
        .versionId(versionId)
        .creationDate(creationDate)
        .caCertPath(caCertPath)
        .authPrivateKeyPath(authPrivateKeyPath)
        .build();
  }

  public static Builder builder() {
    return new AutoValue_FlowerSupernodeContainerConfig.Builder();
  }

  public abstract Builder toBuilder();

  @AutoValue.Builder
  public abstract static class Builder extends AbstractFlowerContainerConfig.Builder<Builder> {

    public abstract Builder caCertPath(@Nullable String caCertPath);

    public abstract Builder authPrivateKeyPath(@Nullable String authPrivateKeyPath);

    @Nullable
    abstract String getCaCertPath();

    @Nullable
    abstract String getAuthPrivateKeyPath();

    abstract FlowerSupernodeContainerConfig autoBuild();

    public FlowerSupernodeContainerConfig build() {
      if (getCaCertPath() == null) caCertPath("data/system/flower/ca.crt");
      if (getAuthPrivateKeyPath() == null) authPrivateKeyPath("data/system/flower/credentials");
      return autoBuild();
    }
  }
}
```

FlowerSuperexecContainerConfig follows the same pattern — extend AbstractFlowerContainerConfig, keep only getFabWhitelistPath(), getType(), create(...), and a Builder extends AbstractFlowerContainerConfig.Builder<Builder> with the fabWhitelistPath setter/getter and the getName() abstract getter it needs for the default-path logic in build().

A few things worth knowing before you do this:

AutoValue only allows one @AutoValue class per hierarchy branch — you can't make AbstractFlowerContainerConfig itself @AutoValue and then extend it with another @AutoValue class. It has to be a plain abstract class, which is what's shown above.
Inherited abstract getters/setters are picked up automatically — you don't need to redeclare getName(), host(...), etc. in the subclasses; AutoValue's annotation processor walks the full supertype hierarchy.
The generic self-typed builder (Builder<B extends Builder<B>>) is what lets the common setters return the correct concrete builder type for chaining (.name(...).image(...).caCertPath(...)). This is a known, working AutoValue idiom, but it does add a layer of generics indirection that's slightly less readable than flat classes.
Given you only have two classes sharing this, it's a judgment call — this removes real duplication (11 properties × getter+setter, twice), but introduces generics and an extra file. If you expect more Flower/container variants down the line, it pays off quickly; if these two are likely to stay as the only ones, the current duplication might honestly be fine.

marikaris (changes requested):

armadillo/src/main/java/org/molgenis/armadillo/container/DockerService.java
timcadman: As in do we want the URL settable in the UI? I think if so maybe in the flower container UI page rather than system? But yes you're right, then it would need API. We can also start off leaving in application.yml as we build a MVP and then expand later?

armadillo/src/main/java/org/molgenis/armadillo/container/FlowerSuperexecContainerConfig.java
timcadman: Yes exactly, its needed to keep ContainerConfig interface happy

armadillo/src/main/java/org/molgenis/armadillo/container/DockerService.java
marikaris: I think we can work out the details later and it should be a story to pick up later for sure. Just good to keep in mind, I think :)

armadillo/src/main/java/org/molgenis/armadillo/container/FlowerSuperexecContainerConfig.java
marikaris: Ok!

marikaris (approved):

PR #1014, opened 2026-04-10
test: fix expected error to work for rock and rserver

timcadman:
## Background
There is one resource test on which both rserve and rock give the same behaviour (error message), but the content of the error message differs, which makes the test fail for rserver.

## What's changed
Broaden the regex so the test passes on failure with either error message

## How to test
- [ ] Run release test on Armadillo for `donkey` profile
- [ ] Run release test on Armadillo for `socker` profile (`datashield/rserver_soccer-donkey:latest`)
- [ ] Check CI green

### DIFF
--- a/scripts/release/testthat/tests/test-10-resources.R
+++ b/scripts/release/testthat/tests/test-10-resources.R
@@ -94,7 +94,7 @@ test_that("/objects/ (iii): resource file accessible, data inaccessible - assign
       symbol = "old_iii_resolved",
       expr = as.symbol("as.resource.object(old_iii)")
     ),
-    "invalid first argument",
+    "invalid first argument|valor ausente",
     info = "/objects/: researcher must have access to data project to resolve"
   )
 })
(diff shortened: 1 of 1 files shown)


marikaris (changes requested):
Doesn't work on my laptop:
```
Failure (test-10-resources.R:91:3): /objects/ (iii): resource file accessible, data inaccessible - assign succeeds, resolve fails
`datashield.assign.expr(...)` threw an error with unexpected message.
Expected match: "invalid first argument|valor ausente"
Actual message: "\033[1m\033[22mThere are some DataSHIELD errors:\n\033[31m✖\033[39m \033[1m\033[33marmadillo\033[39m: missing value where TRUE/FALSE needed\033[22m\n"
/objects/: researcher must have access to data project to resolve
Backtrace:
    ▆
 1. ├─testthat::expect_error(...) at test-10-resources.R:91:3
 2. │ └─testthat:::quasi_capture(...)
 3. │   ├─testthat (local) .capture(...)
 4. │   │ └─base::withCallingHandlers(...)
 5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 6. └─DSI::datashield.assign.expr(...)
 7.   └─DSI:::.handle_errors(errors.print)
 8.     └─cli::cli_abort(...)
 9.       └─rlang::abort(...)
```

marikaris (approved):

PR #1015, opened 2026-04-10
feat: login backoff

timcadman:
## Background
The pen test identified as a potential security vulnerability the fact that people could submit unlimited failed login attempts with basic auth, opening the door to brute-force attacks.

## What's changed
Now implemented a back-off: users get 5 'free' attempts, after which there is a delay of 1 minute before they can login again, increasing exponentially to a maximum of 30 minutes.

## How to test
- [ ] Code review
- [ ] UI review
- [ ] Check CI green


LOGIN_BACKOFF_PLAN.md
marikaris, line 1: I think this would be nice to move to our documentation as well. In my OIDC UI PR, I moved some things around in the documentation so that we have a folder for developer documentation. I think it would fit there.

marikaris (commented):
Works nicely. Only suggestion to move docs to actual documentation. See comment.

marikaris:
Just one more thing: would it be possible to show how many attempts a user has left?

timcadman:
Done - now gives a warning on incorrect password entries prior to lockout. Test by entering wrong password in the UI and see message.

marikaris (approved):

PR #1017, opened 2026-04-17
feat: update armadillo in UI

marikaris:
# How to test

## Setup test environment
**IMPORTANT: Will only work on MacOS with sudo rights**
Add the following to your `application.yml`:
```
armadillo:
  armadillo-home: /path/to/molgenis-service-armadillo
  armadillo-config-file: /path/to/application.yml
  armadillo-mode: DEV
```

1. Create config file for running armadillo globally on your macbook:
`sudo nano /Library/LaunchAgents/org.molgenis.armadillo.plist`
Add the following (replace `{ARMADILLO_LOCATION}` with the root of your armadillo clone):
```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
    <dict>
        <key>Label</key>
        <string>org.molgenis.armadillo</string>

        <key>ProgramArguments</key>
        <array>
            <string>/usr/bin/java</string>
            <string>-jar</string>
            <string>{ARMADILLO_LOCATION}/armadillo.jar</string>
        </array>
        <key>RunAtLoad</key>
        <true/>
        <key>KeepAlive</key>
        <true/> <!-- run the program again if it terminates -->
        <key>WorkingDirectory</key>
        <string>{ARMADILLO_LOCATION}/</string>
        <key>StandardErrorPath</key>
        <string>{ARMADILLO_LOCATION}/logs/armadillo-global.err</string>
        <key>StandardOutPath</key>
        <string>{ARMADILLO_LOCATION}/logs/armadillo-global.out</string>
    </dict>
</plist>
```
2. Gradle build armadillo on this PR branch (assuming this will land in `/build/libs`)
3. In the root of your armadillo clone (as set in previous config), do the following:
(check if version in command below is correct according to your build)
```ln -s build/libs/molgenis-armadillo-5.14.0-SNAPSHOT.jar armadillo.jar```
4. Enable the global run of armadillo:
```bash
loggedInUser=$( ls -l /dev/console | awk '{print $3}' )
userID=$( id -u $loggedInUser )
launchctl enable gui/${userID}/org.molgenis.armadillo.plist
```
5. Run in root of armadillo clone:
``` bash
cp scripts/install/armadillo-update.sh .
chmod a+x armadillo-update.sh
```
6. Start up global armadillo service:
```launchctl bootstrap gui/${userID} /Library/LaunchAgents/org.molgenis.armadillo.plist```
7. Go to localhost:8080 to see if armadillo is running
8. To get log, run in root of armadillo: 
`tail -1000f /logs/armadillo-global.out`
Logs of the script after bringing armadillo down via update, can be found in: `logs/update.log`. This can't be put in the armadillo log, as armadillo is being brought down and up again, losing this information.

### Test plan for UI
- [x] Go to UI, login with oidc user as admin (to verify auth server config works correctly)
- [x] Click on System in the menu bar, select "Control"
- [ ] See that there's a new version available, this will be the latest release (5.12.2)
- [ ] Click on download
- [ ] See that progressbar shows while downloading 
- [x] When download is done, bar is green, success message shows
- [x] Test if other version can be downloaded by specifying version number (for example 5.13.1)
- [x] See that all downloaded versions are in the dropdown
- [x] Scroll down and see the releases you have just downloaded are in the list of jars
- [x] See that current version is disabled and cannot be deleted
- [x] Delete the other version you downloaded
- [x] Edit the "Auth server" 
- [x] Fill in some things, click on cancel
- [x] See that state is reset
- [x] Edit again, fill in information of other auth server (when using fusionauth, switch to keycloak and other way around)
- [ ] Click on save
- [ ] See warning message, information sufficient?
- [ ] Click yes
- [ ] Refresh
- [ ]  Log in using oidc, see different auth server
- [ ] Go back to control
- [ ] Press soft restart
- [ ]  Refresh, log back in, back to control page
- [ ] Press hard restart
- [ ] Refresh, log back in, back to control page
- [ ] Now finally press the update button
- [ ] Read warning message, sufficient?
- [ ] Click yes
- [ ] Refresh, see version changed to 5.12.2

### Test for production
- [ ] Install on demo server and execute above testplan

### Other
- [ ] Documentation sufficient

todo:
- [x] documentation on configuring keycloak
- [x] documentation on running update feature on dev
- [x] documentation on update feature (python required)
- [x] rebuild documentation
- [x] added tests
- [x] Restart endpoint
- [x] endpoint for updating oidc config (backups application.yml and adds new props no new one then restarts application)
- [x] Download latest armadillo endpoint
- [x] Download specific version edpoint
- [x] Endpoint for checking if update available
- [x] Endpoint for deleting jars
- [x] Endpoint for listing available jars
- [x] Script for bringing armadillo down, creating symlink and restarting
- [x] Update endpoint that triggers script after doing prep
- [x] UI
- [x] Disable when running in Docker container
- [x] hide armadillo.jar for delete
- [x] look into auth server uri/client id/opaque id same??? bug???
- [x] rollback application.yml if changed and not getting back up, rollback appliation if update and not getting back up
- [x] Get working keycloak config for demo server
- [x] test for get pid
- [x] UI tests

### CODE REVIEW

**Comment by timcadman** [comment]
## Test on demo notes
- "Click on download" - there is no 'download' option, it just says 'update now'
- "Test if other version can be downloaded by specifying version number (for example 5.13.1)" - I tried this, clicked 'download' and nothing happened
- Error: Cannot update OIDC config, because: ApiError: java.nio.file.AccessDeniedException: /etc/armadillo/application.yml.bak.

File: scripts/install/armadillo-reboot.sh [hunk 1]
```diff
@@ -0,0 +1,169 @@
...
+
+increase_timeout() {
+  TIMEOUT=$(( TIMEOUT * 2))
+}
+
+restart_if_down() {
+  sleep $TIMEOUT
+  # check if server up
+  echo "👩‍🔬 Checking if everything went correctly and if Armadillo is up and running 🏃‍➡️..."
+  SERVER_UP="$(lsof -i :8080 | grep java)"
```
DickPostma, line 86: Port 8080 is configurable, so if end users changing this port this script gives a false positive. Maybe check if $PID is still alive?

scripts/install/armadillo-reboot.sh
```diff
@@ -0,0 +1,169 @@
...
+    else
+      # default build path when running with gradle/intellij
+      BUILD_DIR="$ARMADILLO_PATH/build/libs"
+      # macos
+      OLD_JAR=$(stat -f %Y armadillo.jar)
+  fi
+  # replace v if vx.y.z pattern is used for specifying version
+  JAR_NAME="molgenis-armadillo-${ARMADILLO_VERSION/v/}.jar"
+  VERSION_FOUND=$( check_version_exists "$BUILD_DIR" "$JAR_NAME" "$ARMADILLO_VERSION" )
+  if [[ $VERSION_FOUND ]];
```
DickPostma: $VERSION_FOUND is returning a string if true.. (cause off the echo in the function)

scripts/install/armadillo-reboot.sh
```diff
@@ -0,0 +1,169 @@
...
+    else
+      echo "❌ ERROR: No jar available for version $ARMADILLO_VERSION. Please download it."
+      exit_script
+  fi
+  else
+    echo "Nothing fancy, just a restart"
+    restart_armadillo ""
+fi
+
+increase_timeout() {
```
DickPostma, line 68: Maybe handy to stop after x seconds? Cause I think with an error this is gonna be an infinite loop?

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
```diff
@@ -0,0 +1,192 @@
...
+  public static final String ARMADILLO_VERSION = "ARMADILLO_VERSION";
+  private final ManagementService managementService;
+  private final AuditEventPublisher auditor;
+
+  public ManagementController(ManagementService managementService, AuditEventPublisher auditor) {
+    this.managementService = requireNonNull(managementService);
+    this.auditor = auditor;
+  }
+
+  @Operation(summary = "Soft restart armadillo")
```
jhhaanstra: Specify the definitions of soft and hard reset in the `@Operation` summaries

armadillo/src/main/java/org/molgenis/armadillo/config/ApplicationConfigUpdater.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  String transformLine(String line, ConfigParseState state, OidcDetails oidcDetails) {
+    if (line.strip().startsWith("#")) return line;
+
+    updateState(line, state);
+    return applyTransformation(line, state, oidcDetails);
+  }
+
+  void updateState(String line, ConfigParseState state) {
```
jhhaanstra: This can be done a lot simpler using a YAML parser like SnakeYAML: https://www.baeldung.com/java-snake-yaml

scripts/install/armadillo-reboot.sh
marikaris: Hmm ok, but this is checking if armadillo is back up, so I don't know what the pid will be as it won't be running on the same PID as the process we killed. Is there any other way to check?

scripts/install/armadillo-reboot.sh
marikaris: Thanks! I'm pretty sure I had it in there at some point, but it might have gotten lost when moving things around. I'm pretty sure Bart said this to me when I demo'ed it a while back. It's a good point though, I'll fix it :)

armadillo/src/main/java/org/molgenis/armadillo/config/ApplicationConfigUpdater.java
```diff
@@ -0,0 +1,97 @@
+package org.molgenis.armadillo.config;
+
+import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.LinkedHashMap;
+import org.molgenis.armadillo.metadata.OidcDetails;
+import org.yaml.snakeyaml.Yaml;
+
+public class ApplicationConfigUpdater {
```
jhhaanstra: I have two major comments for this file:

### 1. Make an internal model for the config
The current type for a config is a very deeply-nested hashmap:
```java
LinkedHashMap<
          String,
          LinkedHashMap<
              String,
              LinkedHashMap<
                  String,
                  LinkedHashMap<
                      String,
                      LinkedHashMap<
                          String, LinkedHashMap<String, LinkedHashMap<String, Object>>>>>>>
```
I'd suggest creating a POJO model that you use instead and using that to access everything. It's more code, but it creates a clearer image of how the config is structured and what everything means. I have Claude generate the following snippet: https://gist.github.com/jhhaanstra/faad370a87f12ffa7abcb5c7f1f85b6d



### 2. Make the class immutable
Instead of making all fields mutable, it's maybe nicer to have all fields as `private final`. Make the field that you access outside of the class, which is just the config, available through a getter.

I think the updater should have the following flow:
- Create the object
```java
update = new ApplicationConfigUpdater(configFilePath);
```
- Then inside the constructor you can read the config file immediately when creating the object.
- You can access the config via a getter-method:
```java
public Config getConfig() {
   return config;
}
```

armadillo/src/main/java/org/molgenis/armadillo/config/ApplicationConfigUpdater.java
```diff
@@ -0,0 +1,97 @@
...
+
+    providerMolgenisConfig.put(issuerUri, oidcDetails.getIssuerUri());
+    registrationMolgenisConfig.put(clientId, oidcDetails.getClientId());
+    registrationMolgenisConfig.put("client-secret", oidcDetails.getClientSecret());
+    opaquetokenConfig.put(clientId, oidcDetails.getDeviceClientId());
+    jwtConfig.put(issuerUri, oidcDetails.getDeviceIssuerUri());
+  }
+
+  void writeConfigFile(String path) throws IOException {
+    Yaml yaml = new Yaml();
```
jhhaanstra: I think this can be a `private static final`

armadillo/src/main/java/org/molgenis/armadillo/controller/LoginController.java
```diff
@@ -0,0 +1,30 @@
...
+@Validated
+@Hidden
+public class LoginController {
+  @GetMapping("/oauth2/")
+  public RedirectView whenAuthenticatedRedirect() {
+    return new RedirectView("/");
+  }
+
+  @GetMapping("/basic-login")
+  public RedirectView basicLogin(Principal principal) {
```
jhhaanstra: Principal is never used here

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
```diff
@@ -0,0 +1,192 @@
...
+
+@Tag(name = "manage", description = "Manage the application and settings")
+@RestController
+@SecurityRequirement(name = "http")
+@SecurityRequirement(name = "bearerAuth")
+@SecurityRequirement(name = "JSESSIONID")
+@RequestMapping("manage")
+@PreAuthorize("hasRole('ROLE_SU')")
+public class ManagementController {
+  public static final String ARMADILLO_VERSION = "ARMADILLO_VERSION";
```
jhhaanstra: This can be private, as it's never used outside of this class

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
```diff
@@ -0,0 +1,192 @@
...
+        Map.of("VERSION_TO_DELETE", version));
+  }
+
+  @Operation(summary = "Get info of latest release")
+  @GetMapping("app/latest-release-info")
+  public Object getLastReleaseInfo(Principal principal) {
+    return auditor.audit(
+        () -> {
+          try {
+            Gson gson = new Gson();
```
jhhaanstra: Can also be reused, so private static final

armadillo/src/main/java/org/molgenis/armadillo/exceptions/RebootScriptRunFailedException.java
```diff
@@ -0,0 +1,8 @@
+package org.molgenis.armadillo.exceptions;
+
+public class RebootScriptRunFailedException extends RuntimeException {
+
+  public RebootScriptRunFailedException(String message) {
+    super(message);
```
jhhaanstra: At this point you're dropping the root cause of the exception. Wouldn't you want something like this as a constructor?
```java
  public RebootScriptRunFailedException(String message, Exception cause) {
    super("Reboot script failed: " + message, cause);
  }
```

armadillo/src/main/java/org/molgenis/armadillo/interceptor/WebConfig.java
```diff
@@ -45,4 +47,9 @@ public WebServerFactoryCustomizer<TomcatServletWebServerFactory> tomcatCustomize
             connector ->
                 connector.setEncodedSolidusHandling(EncodedSolidusHandling.DECODE.getValue()));
   }
+
+  @Bean
+  public HttpClient httpClient() {
+    return HttpClient.newBuilder().proxy(ProxySelector.getDefault()).build();
```
jhhaanstra: Out of curiosity, why the proxy selection?

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
```diff
@@ -0,0 +1,335 @@
...
+
+  // location of update log
+  private String updateLogPath;
+
+  private final RebootScriptRunner scriptRunner;
+
+  private final ApplicationConfigUpdater appConfigUpdater;
+
+  // Constants
+  String rebootScript = "armadillo-reboot.sh";
```
jhhaanstra: In Java constants are written as follows:

```java
private static final String REBOOT_SCRIPT = "armadille-reboot.sh";
```

You usually place them at the top of the class as well.

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
```diff
@@ -0,0 +1,335 @@
...
+  public void softRestartApplication() {
+    ArmadilloServiceApplication.restart();
+  }
+
+  private String getProcessName() {
+    return ManagementFactory.getRuntimeMXBean().getName();
+  }
+
+  String getJavaProcessId(String processName) {
+    return Arrays.stream(processName.split("@")).toList().getFirst();
```
jhhaanstra, line 102: What if the list is empty? Do we want to take that case into account or is that out of scope?

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
```diff
@@ -0,0 +1,335 @@
...
+      @Value("${stdout.log.path:./logs/armadillo.log}") String logPath,
+      @Value("${update.log.path:#{null}}") String updatePath,
+      @Value("${armadillo.armadillo-config-file:/etc/armadillo/application.yml}")
+          String armadilloConfigFile,
+      @Autowired BuildProperties buildProperties,
+      HttpClient httpClient) {
+    this.httpClient = httpClient;
+    this.buildProperties = buildProperties;
+    if (updatePath == null) {
+      var splittedLogFilepath = logPath.split(Pattern.quote(File.separator));
```
jhhaanstra: Nitpick: move this into a separate method with a good name, makes it more readable

armadillo/src/main/java/org/molgenis/armadillo/service/RebootScriptRunner.java
```diff
@@ -0,0 +1,96 @@
...
+        throw new IOException("File cannot be created");
+      }
+    }
+    return logFile;
+  }
+
+  String createPythonScriptForThread(String pythonList, String absoluteLogFilePath) {
+    String scriptTemplate =
+        """
+                                  import os, sys, subprocess
```
jhhaanstra: Nitpick: align this nicely

armadillo/src/main/java/org/molgenis/armadillo/storage/FileDownloader.java
```diff
@@ -0,0 +1,77 @@
...
+    return (total * 100) / current;
+  }
+
+  static void processFile(
+      FileOutputStream fileOutputStream,
+      BufferedInputStream in,
+      long fileSize,
+      LongConsumer progressCallback)
+      throws IOException {
+    byte[] dataBuffer = new byte[8192];
```
jhhaanstra, line 43: Why this specific array size?

jhhaanstra (changes requested):

jhhaanstra:
This is a cool feature, but very tricky since there are a lot of tricks to work around the JVM. One of my main concerns is: what happens when one piece of the puzzle fails and how is that resolved?

docs/pages/basic_usage/auth.md
```diff
@@ -2,4 +2,109 @@
...
+              - authorization_code
+              - refresh_token
+      resourceserver:
+        jwt:
+          issuer-uri: https://the-auth-server # can be different one
+        opaquetoken:
+          client-id: the-device-flow-client-id
+```
+Note that both the opaquetoken and the client-id are specified here. 
+In FusionAuth the client-id and opaguetoken client-id can be the same one.
```
timcadman: "can be" --> "will be"

docs/pages/basic_usage/auth.md
```diff
@@ -2,4 +2,109 @@
...
+
+After initial configuration, auth server configuration can be adjusted in Armadillo's UI. 
+To do so, navigate to `System > Control`. Scroll to "Authentication server" and fill in the fields. Then save, read
+the message carefully and press "Yes". The server will restart with the new configuration applied.
+
+
+![auth_server.png](../../img/ui/auth_server.png)
+
+## Setup in FusionAuth
+To configure a client in FusionAuth, add a configuration in Applications on your auth server.
```
timcadman: "on your auth server as follows"

docs/pages/basic_usage/application_update_restart.md
```diff
@@ -0,0 +1,81 @@
...
+administrator for help.
+
+### Updating to a different version
+If you want to update to a different version than the last stable one that will show up in the update alert, you can 
+click the button next to Advanced update to reveal the advanced update features: 
+
+![armadillo_custom_update.png](../../img/ui/armadillo_custom_update.png)
+
+Here any tag available in 
+[the armadillo github repository](https://github.com/molgenis/molgenis-service-armadillo/releases) .
```
timcadman: .. can be downloaded.

docs/pages/basic_usage/application_update_restart.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Storage
+Because downloading Armadillo versions can fill up the server, it is possible to delete downloaded versions that are not
+in use. To do so, scroll to the "Storage" section. Select the version you want to delete and push the delete button. 
+
+![armadillo_delete_jar.png](../../img/ui/armadillo_delete_jar.png)
+
+## Authentication server
+The last section on the Control page is "Authentication Server". Information about this can be found in the 
+[Authentication server section](http://127.0.0.1:8000/molgenis-service-armadillo/pages/basic_usage/auth/#ui) of this
```
timcadman: Should it be relative link rather than localhost?

docs/pages/basic_usage/application_update_restart.md
```diff
@@ -0,0 +1,81 @@
...
+## Restart
+Within the control screen, it's also possible to restart the armadillo server. There are two options: 
+
+- Soft restart
+- Hard restart
+
+### Soft restart
+This will attempt to restart within the application itself, without truly bringing it down. This button works even if 
+python is not installed on your armadillo. Because armadillo is never truly brought down, risks of it not coming back 
+up are slightly decreased. In the very unlike scenario that Armadillo doesn't come back up, contact your administrator.
```
timcadman: unlikely

docs/pages/basic_usage/application_update_restart.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Hard restart
+This will trigger a full restart of the application, calling a script that will bring armadillo down and later back up. 
+If this process fails, the script will keep attempting to restart 10 times after increasing periods of time 
+(4 seconds, 8 seconds, 16 seconds, etc.). In the very unlike scenario that Armadillo doesn't come back up, contact your
```
timcadman: unlikely

docs/pages/basic_usage/auth.md
timcadman: opaque

docs/pages/basic_usage/auth.md
```diff
@@ -2,4 +2,109 @@
...
+Create a new client to use as regular oidc client. Use the Client ID you configure in here as `molgenis.client-id` in 
+above setup in the `application.yml`. For the secret, go to the Credentials tab and get the Client Secret there. Apply
+the settings below.
+
+#### Settings 
+- Set Root, Home URLs, Web origins and Admin URL to the URL of your armadillo server
+- Set valid redirect URLs: http://your-armadillo-url/login/oauth2/code/molgenis
+- Set Valid post logout redirect URIs:  http://your-armadillo-url/logout
+- Client authentication set to On
+- Autorization set to off
```
timcadman: Authorization

docs/pages/basic_usage/index.md
```diff
@@ -10,8 +10,37 @@ Depending on your [role](../quick_start.md#quick-start) you have different optio
...
+        <span style="background-color:#D4ECFF; padding: 5px 0px 2.5px 5px; border-radius: 3px 0px 0px 3px; ">
+            <svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" fill="#353535" class="bi bi-tag" viewBox="0 0 16 16">
+                <path d="M6 4.5a1.5 1.5 0 1 1-3 0 1.5 1.5 0 0 1 3 0m-1 0a.5.5 0 1 0-1 0 .5.5 0 0 0 1 0"/>
+                <path d="M2 1h4.586a1 1 0 0 1 .707.293l7 7a1 1 0 0 1 0 1.414l-4.586 4.586a1 1 0 0 1-1.414 0l-7-7A1 1 0 0 1 1 6.586V2a1 1 0 0 1 1-1m0 5.586 7 7L13.586 9l-7-7H2z"/>
+            </svg>
+        </span>
+        <span style="margin: 0 0 0.5em 0; border-left: 3px solid #017FFD; padding: 5px 0 3px 0.5em">5.14.0</span>
+    </div>
+
+    To ease some of the system administrators responsibilites, datamanagers can access the Armadillo Control page in the User Interface where they can:
```
timcadman: administrator's and data managers

docs/pages/dev_guides.md
```diff
@@ -94,7 +94,7 @@ information to get you started and help you get familiarised with our code and w
     For testing without having to installing Java you can run using docker:
 
     1. Install [docker-compose](https://docs.docker.com/compose/install/)
-    2. Download this [docker-compose.yml](docker-compose.yml).
+    2. Download this [https://github.com/molgenis/molgenis-service-armadillo/blob/master/docker-compose.yml](docker-compose.yml).
```
timcadman: Should be [docker-compose.yml](https://github.com/molgenis/molgenis-service-armadillo/blob/master/docker-compose.yml)?

docs/pages/install_management/armadillo_install.md
```diff
@@ -21,6 +21,7 @@ In case of using dsOmics this setup can be rather bigger. Please [contact](../co
 
 * Java 21 JRE or JDK
 * Docker
+* Python 3 (to enable auth server reconfiguration and updating capabilities in the UI )
```
timcadman: Space between UI and )

timcadman (changes requested):

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
```diff
@@ -0,0 +1,192 @@
...
+            throw new ResponseStatusException(HttpStatus.BAD_REQUEST, e.getMessage());
+          }
+        },
+        principal,
+        "UPDATE_ARMADILLO",
+        Map.of(ARMADILLO_VERSION, version));
+  }
+
+  @Operation(summary = "List all available jars")
+  @GetMapping("app/list")
```
jhhaanstra, line 102: What is the process of making sure that the disk doesn't fill up with unused jars again? And how do we communicate to the user that this is something that can happen here?

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
```diff
@@ -0,0 +1,192 @@
...
+  public Set<String> listAvailable(Principal principal) {
+    return auditor.audit(
+        managementService::listAvailableJars, principal, "LIST_AVAILABLE_VERSIONS");
+  }
+
+  @Operation(summary = "Get current OIDC config")
+  @GetMapping("auth/oidc-config")
+  public Map<String, String> getOidcConfig(Principal principal) {
+    return auditor.audit(managementService::getCurrentOidcConfig, principal, "GET_OIDC_CONFIG");
```
jhhaanstra: A point of discussion: this exposes the OIDC client secret, which is sensitive security information. 
"In NHI security, the client secret sits in the same risk class as other long-lived secrets because compromise enables token theft, session hijacking, and impersonation of the application." - https://nhimg.org/glossary/oidc-client-secret/
What's the risk that we're running here, and is this acceptable?

armadillo/src/main/java/org/molgenis/armadillo/exceptions/RebootScriptRunFailedException.java
jhhaanstra, line 1: You can consider moving this to that RebootScriptRunner and make it an inner class, as that's the only place that this exception is created, making it a bit more exclusive to that scope (like how we've done that with `ApplicationConfigUpdater.ConfigUpdateException`). But I'll leave that up to you, I'm not surewhat the code conventionts of this repo are either..

armadillo/src/main/java/org/molgenis/armadillo/service/RebootScriptRunner.java
```diff
@@ -0,0 +1,96 @@
...
+                                  if os.fork() > 0:
+                                      sys.exit(0)
+                                  os.setsid()
+                                  if os.fork() > 0:
+                                      sys.exit(0)
+                                  with open('%s', 'a') as log:
+                                      subprocess.run(%s, stdout=log, stderr=log, stdin=subprocess.DEVNULL)
+                                  """;
+    return String.format(scriptTemplate, absoluteLogFilePath, pythonList);
```
jhhaanstra: You can do this in 1 line using:
```java
    return """
           import os, sys, subprocess
           if os.fork() > 0:
               sys.exit(0)
           os.setsid()
           if os.fork() > 0:
               sys.exit(0)
           with open('%s', 'a') as log:
               subprocess.run(%s, stdout=log, stderr=log, stdin=subprocess.DEVNULL)
           """
        .formatted(absoluteLogFilePath, pythonList);
```

armadillo/src/test/java/org/molgenis/armadillo/service/RebootScriptRunnerTest.java
jhhaanstra, line 1: maybe an additional test would be to check whether the scriptrunner actually runs the python script by verifying the outlput log? Say that the script is `print('test')`, wouldn't the output log file not simply contain the string test?

armadillo/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
     testImplementation 'org.springframework.boot:spring-boot-starter-test'
     testImplementation 'org.springframework.security:spring-security-test'
     testImplementation "junit:junit"
+    testImplementation("org.awaitility:awaitility:4.3.0")
```
jhhaanstra, line 45: awaility is used nowhere right? So I think this dependency is redundant? Or did I overlook something?

armadillo/src/test/java/org/molgenis/armadillo/service/RebootScriptRunnerTest.java
```diff
@@ -0,0 +1,115 @@
...
+    m.setAccessible(true);
+
+    String result =
+        (String) m.invoke(scriptRunner, (Object) new String[] {"/path/script", "-p", "/home"});
+    assertEquals("['/path/script', '-p', '/home']", result);
+  }
+
+  @Test
+  void buildPythonCommand_escapesSingleQuotes() throws Exception {
+    Method m = RebootScriptRunner.class.getDeclaredMethod("buildPythonCommand", String[].class);
```
jhhaanstra: Isn't that fact that this method needs to be hacked and made accessible a sign that we need to add a dependency on a `ScriptProvider` interface or something? Then you can test that exclusively instead of having to work around all the Reboot logic

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
marikaris: Yeah, I've had this discussion with Dick. I find it a bit difficult. I understand the problem, but if you're getting this information from the getOidc you already have SUDO rights: `@PreAuthorize("hasRole('ROLE_SU')")` on `ManagementController`.
That would mean that nothing would stop you from changing the OIDC config to anything else (even your own auth server). While it would be possible to remove the secret from the get endpoint, that would make the experience a bit weird for users and I don't think having that get will make it more unsafe than having this whole feature already is. Or am I missing something?

armadillo/src/main/java/org/molgenis/armadillo/controller/ManagementController.java
marikaris: Image

This is shown in the UI, so storage on the server is presented above the list of jars. Do you think that's sufficient? Jars will only be downloaded when users click the "download" button, so it won't be filled up automatically. Do you think this is sufficient?

armadillo/src/main/java/org/molgenis/armadillo/interceptor/WebConfig.java
marikaris: I honestly don't remember. I added this bean to fix something I broke while refactoring to make testing easier. I found the stackoverflow post I used to help fixing this:
https://stackoverflow.com/questions/56791694/how-to-enable-httpclient-for-springboot-2-0-2

That one however doesn't refer to the proxy, so did some more digging. I think I figured the downloadFile function (now in FileDownloader class) already used the httpClient to retrieve the releases from github. That piece of code has been in there since april 13th and I can't go back that far in my browsing history, so I'm not sure where I got that form. 

So long answer short: I don't entirely know, just that it works. Would you do it differently?

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
```diff
@@ -0,0 +1,317 @@
...
+  private static final String DOWNLOAD_COMPLETE = "Download complete";
+
+  @Value("${armadillo.armadillo-home:/usr/share/armadillo/application}")
+  String armadilloHome;
+
+  @Value("${armadillo.armadillo-mode:PROD}")
+  String armadilloMode;
+
+  @Value("${spring.security.oauth2.client.registration.molgenis.client-id:#{null}}")
+  String clientId;
```
jhhaanstra: Instead of a lot of the separate fields, maybe you could use an OidcDetails object instead? You can inject that via `AppConfig` and use these `@Value` annotated fields as params for the `@Bean` method?

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
```diff
@@ -0,0 +1,317 @@
...
+    currentConfig.put("issuerUri", issuerUri);
+    currentConfig.put("clientId", clientId);
+    currentConfig.put("clientSecret", clientSecret);
+    currentConfig.put("deviceClientId", deviceClientId);
+    currentConfig.put("deviceIssuerUri", deviceIssuerUri);
+    return currentConfig;
+  }
+
+  private String getScriptVersionTag(String version) {
+    // TODO: simplify this after update script is available on version, then it should simply return
```
jhhaanstra: Is this outside scope for now? It's good practice to not use `TODO` in code, but create an issue on the project scrum board instead. That way it'll be easier to track, and it'll be more dynamic because people can comment on it for example. 
Finally, these comments can become outdated, and in the future, people don't dare to touch it because they're not sure whether it's actually resolved.

armadillo/src/main/java/org/molgenis/armadillo/ArmadilloServiceApplication.java
```diff
@@ -1,14 +1,39 @@
...
   public static void main(String[] args) {
     SpringApplication.run(ArmadilloServiceApplication.class, args);
   }
+
+  @Override
+  public void setApplicationContext(ApplicationContext context) {
+    ArmadilloServiceApplication.context = (ConfigurableApplicationContext) context;
+  }
+
+  public static void restart() {
```
jhhaanstra, line 21: I think it's good to add some context via a comment here?

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
marikaris: Good plan!

marikaris: https://github.com/molgenis/GCC/issues/3375

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
marikaris: I have moved it to the constructor so I can still test the behaviour in the ManagementService unittest

armadillo/src/main/java/org/molgenis/armadillo/service/ManagementService.java
marikaris: I don't think that the list can be empty. processName cannot be null, it's a string, when we split it, the list will always have 1 item. 
This is being put in the method:
```
java.lang.management.ManagementFactory
public static java.lang.management.RuntimeMXBean getRuntimeMXBean()
 
Returns the managed bean for the runtime system of the Java virtual machine.
```
 Or am I missing something? 

I don't think that will ever be empty as the application is running and thus must have an id?
I tried throwing an exception when the list is empty, but failed to write a test for it, so I think it's very unlikely for this to fail

armadillo/src/main/java/org/molgenis/armadillo/storage/FileDownloader.java
marikaris: https://stackoverflow.com/questions/1111661/8192-bytes-when-creating-file

marikaris: Had to choose something. Not too big, as then t he progress bar won't be updating enough, not too small, because it will slow down the application.

armadillo/src/main/java/org/molgenis/armadillo/exceptions/RebootScriptRunFailedException.java
marikaris: I think I'll leave it. I'm fine with this.

armadillo/src/main/java/org/molgenis/armadillo/storage/FileDownloader.java
jhhaanstra: Ah, I get it now. This is the size of the buffer before writing it to the downlaoded file, causing the progress bar to continue on. I think it would be nice to add some javadoc for this to clarify

armadillo/src/test/java/org/molgenis/armadillo/config/ApplicationConfigFileTest.java
```diff
@@ -0,0 +1,278 @@
...
+    Yaml yaml = new Yaml();
+    Map<String, Object> written;
+    try (InputStream in = Files.newInputStream(outputFile)) {
+      written = yaml.load(in);
+    }
+    assertNotNull(written);
+    assertTrue(written.containsKey("spring"));
+  }
+
+  // ---------- updateApplicationConfig (full round trip) ----------
```
jhhaanstra: These kinds of comments don't really add anything and can become outdated pretty quickly (e.g. you reorder the tests, and there's no check to see whether they're place correctly).
I see this in more places in the PR. Please go over the PR and remove this. If I run all tests underneath this comment, I get the impression I have tested everything that has to do with updating the application config, but reordering once and this not being updated properly, can give me a false sense of security.

armadillo/build.gradle
jhhaanstra: We use it now :)

armadillo/src/main/java/org/molgenis/armadillo/storage/DefaultJarDownloader.java
```diff
@@ -0,0 +1,18 @@
+package org.molgenis.armadillo.storage;
+
+import java.util.function.LongConsumer;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DefaultJarDownloader implements JarDownloader {
```
jhhaanstra, line 23: Why not make FileDownloader implement the interface instead of delegating with an extra class on top? Seems a bit unneeded to me. I think we already discussed this though and you had a good reason for this 😓

erikzwart:
Functional test on https://armadillo-demo.molgenis.net/, first step was to give user (myself) admin rights, after that I followed the steps below:

Test plan for UI

- [x]  Go to UI, login with oidc user as admin (to verify auth server config works correctly)
- [x]  Click on System in the menu bar, select "Control"
- [x]  See that there's a new version available, this will be the latest release (5.12.2)
- [ ]  Click on download (Instead was called update?)

After clicking to download the latest version I did **not** see a progress bar and system seemed to be upgraded to version 5.14.0. I reloaded the page and was still logged in (OIDC).
I was unable to follow the rest of the test steps because I do not seem to have admin rights, however I do see the `control` page:
image

The `Users` page is empty:
image

marikaris:
moved to other PR (with rebase)

PR #1019, opened 2026-05-20
Fix keycloak Device-Flow login

DickPostma:
Device Flow login needed some changes and and audience for the new armadillo.login function.


RubenOPS (approved):
👍

PR #1029, opened 2026-06-25
feat: leading zeros string

### LINKED ISSUE #1027: feat: store columns with leading zeros as string rather than integer
### How to Reproduce
- Create an example project and folder in the UI, e.g. `testnum/data`
- Upload a .csv file via the UI where a column contains numbers with leading zeros, e.g. 

```
id,zip,age
001,00501,34
002,01010,57
003,02134,29
004,00902,41
005,07310,63
006,09001,22
007,06450,48
008,00601,39
009,08402,51
010,01913,45
```
- Download that file in R
```
library(MolgenisArmadillo)
armadillo.login("http://localhost:8080/")
armadillo.load_table("testnum", "data", test_file")
```
- Preview the file in R.

### Expected behavior
Zeros are preserved (e.g. participant ID, postcode ... my postcode here is 08028)
This is debatable though - maybe this behaviour is correct and the solution is the column type review feature request.

### Observed behavior
001, 002, 003 --> 1, 2, 3

timcadman:
## Background

Columns containing leading zeros (e.g. participant IDs `001`, `002`, or postcodes like   `08028`) were being inferred as integers during CSV→Parquet conversion. On write, `Integer.parseInt("001")` returns `1`, so the leading zeros were permanently lost in the stored Parquet file — when the data was downloaded and loaded in R, `001` came back as `1`. Fixes #1027.

## What's changed
- `CharacterSeparatedFile.getTypeOfCell` now classifies an integer-looking value with a significant leading zero as `string` instead of `int`, so the column is stored as a string and the zeros survive the round-trip.
- Added a `hasLeadingZero` helper: strips an optional `+`/`-` sign, then treats a `0` followed by at least one more digit as significant (`"0"` → false, `"00"`/`"001"`/`"-007"` → true). Plain ints (`"0"`, `"5"`) and decimals (`"0.5"`, handled by the    `double` branch) are unaffected.
- The existing `getTypeToSet` logic already makes `string` absorbing, so a column mixing `1` and `002` resolves to `string` for the whole column — no per-cell zero loss.
- Tests in `CharacterSeparatedFileTest` for: leading-zero classification (incl. signed), `"0"` staying `int`, and constructor-level checks for a pure and a mixed leading-zero column.

## How to test
- [ ] Code review
- [ ] Check unit tests pass
- [ ] Manual test:

1. Create new project `csv_zeros`
2. Create folder `data`
3. Save this as a .csv (`test_zeros.csv`)

```id,age
001,30
002,25
00501,40
```
Login via R and download the table:

```
library(MolgenisArmadillo)
armadillo.load_table("csv_zeros", "data", "test_zeros.csv")
```
Check that the `id` column has the leading zeros intact.

### DIFF
--- a/armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
+++ b/armadillo/src/main/java/org/molgenis/armadillo/storage/CharacterSeparatedFile.java
@@ -118,7 +118,7 @@ public class CharacterSeparatedFile {
     try {
       double d = Double.parseDouble(cell);
       if ((d == Math.floor(d)) && !Double.isInfinite(d) && !cell.contains(".")) {
-        return CharacterSeparatedFile.INT;
+        return hasLeadingZero(cell) ? CharacterSeparatedFile.STRING : CharacterSeparatedFile.INT;
       } else {
         return CharacterSeparatedFile.DOUBLE;
       }
@@ -133,6 +133,11 @@ public class CharacterSeparatedFile {
         return CharacterSeparatedFile.STRING;
       }
     }
+  }
+
+  static boolean hasLeadingZero(String cell) {
+    String digits = (cell.startsWith("-") || cell.startsWith("+")) ? cell.substring(1) : cell;
+    return digits.length() > 1 && digits.charAt(0) == '0';
   }
 
   static String getTypeToSet(String type, String currentType) {
--- a/armadillo/src/test/java/org/molgenis/armadillo/storage/CharacterSeparatedFileTest.java
+++ b/armadillo/src/test/java/org/molgenis/armadillo/storage/CharacterSeparatedFileTest.java
@@ -44,6 +44,35 @@ class CharacterSeparatedFileTest {
   }
 
   @Test
+  void testConstructorStoresLeadingZeroColumnAsString() throws IOException, CsvValidationException {
+    String csvData = "id,age\n001,30\n002,25\n";
+    Mockito.when(mockFile.getInputStream())
+        .thenReturn(new ByteArrayInputStream(csvData.getBytes()));
+    Mockito.when(mockFile.getOriginalFilename()).thenReturn("test.csv");
+
+    CharacterSeparatedFile csf = new CharacterSeparatedFile(mockFile);
+
+    assertEquals(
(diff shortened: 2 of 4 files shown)


marikaris:
I'm not really sure how to test, I did the create project and create folder, saved this file as csv:
```
"id","value"
"001",30
"002",25
"00501",40
```
Also tried without header, as you showed. That wouldn't upload. Also tried with and without quotes, but when I upload this, I get an error message when previewing. I tried saving it as csv in excel, but excel does the thing the bug does, removing the 00 in front of the numbers. 

I also don't really get this part:
```
armadillo.load_table("csv_zeros", "data", "test_zeros.csv")
```
The table is supposed to be called `tests_zeros`, right? the csv is confusing me, I'm not sure how I would be able to get it like that in there and if I would, what this fix would have to do with it. Can you just provide me with an actual csv file that I can upload?

marikaris (changes requested):
I managed to test with this csv: 
```
id,name,dob,visits
001,Ana,10/12/2005,3
002,Ben,01/02/1998,7
003,Cara,2024-01-15,1
004,Dan,15-03-1990,12
005,Eve,30/11/2012,0
006,Finn,07/07/2007,5
007,Gail,1985-06-22,9
008,Hugo,21/09/2001,2
009,Ivy,03/03/2033,4
010,Jack,12/12/2012,6
```
It works fine in R, using: `armadillo.load_table("csv-zeros", "data", "names")`. But would it be very hard to fix it in the preview as well?

timcadman:
Fixed the UI. To test: 

1. Upload this .csv (note that 'id' is renamed as 'subjectid' to avoid truncation)
subjectid,name,dob,visits
001,Ana,10/12/2005,3
002,Ben,01/02/1998,7
003,Cara,2024-01-15,1
004,Dan,15-03-1990,12
005,Eve,30/11/2012,0
006,Finn,07/07/2007,5
007,Gail,1985-06-22,9
008,Hugo,21/09/2001,2
009,Ivy,03/03/2033,4
010,Jack,12/12/2012,6

2. Navigate to preview in UI, see that leading zeros are displayed

marikaris (approved):

Issue #902, opened 2025-09-24
As Datamanager I want to be able to upload and download my users workspaces


marikaris:
So that I can download them before deleting, just to make sure it's still stored somewhere locally when a researcher asks for it.

- [ ] Create upload workspace endpoint
- [ ] Create download workspace endpoint
- [ ] Add upload to UI
- [ ] Add download to UI

PR #1045, opened 2026-07-07
feat: upload/download workspace

marikaris:
#902 

how to test:
Start armadillo, open UI
Go to workspaces page
Select a user
Download a workspace
Upload a workspace

todo:
- [x] download endpoint
- [x] tests for download endpoint
- [x] implement upload endpoint
- [x] tests for upload endpoint
- [x] download in UI
- [x] upload in UI
- [x] UI tests


ui/src/views/Workspaces.vue
timcadman, line 1: ## Functional test
### Download workspace: failed
- Workspace name was `armadillo:test_small`
- Chrome gave message that file not available `user-t.j.cadman_at_umcg.nlarmadillo_test_small.RData`
- Armadillo log gives: `Bucket 'user-user-t.j.cadman__at__umcg.nl' doesn't exist, type=org.molgenis.armadillo.exceptions.StorageException, user=user-t.j.cadman__at__umcg.nl, roles=[ROLE_SU]}]`
- Claude diagnosis: 
  - UI sends selectedUser, which is already the folder name: user-t.j.cadman__at__umcg.nl (/all-workspaces keys its map by folder, ArmadilloStorageService.java:243).          
  - Backend runs it through getUserBucketIdentifierFromUserId = "user-" + userId (:269) → user-user-t.j.cadman__at__umcg.nl.                                                   
  - LocalStorageService resolves Paths.get(rootDir, bucket, object) → data/user-user-t.j.cadman__at__umcg.nl/armadillo:test_small.RData.

timcadman, line 1: UI gives success message even when download has failed - should it check? Or not return any message and leave this up to the browser?

timcadman, line 1: ## Functional test
### Upload workspace: failed
1. If I click the box 'select a file to upload' (on a mac), all the files are greyed out and I can't upload anything.
2. I can drag a file to the box to upload, but then I get error `Upload failure: cannot upload workspace for user [t.j.cadman__at__umcg.nl] because ApiError: Not found`. UI console: `Failed to load resource: the server responded with a status of 404 ()`. Trying to hit `workspaces/upload/t.j.cadman__at__umcg.nl`.

timcadman, line 1: If I drag-and-drop a file to the upload box, there is no type guard on the type of file that can be uploaded (e.g. I can upload a .yml file)

timcadman, line 1: When downloading a workspace there is no option to chose the target folder. Should this be setable like with a normal download? How would this work in production, e.g. if I log on to the UI via my local machine, and download a file, is it stored on the server that is running Armadillo or is it stored locally?

timcadman, line 1: Should the workspace size be checked in the UI prior to upload? Or is it ok that this is managed by the backend?

armadillo/src/main/java/org/molgenis/armadillo/controller/DataController.java
```diff
@@ -389,6 +393,17 @@ public List<DSMethod> getAggregateMethods(Principal principal) {
         Map.of());
   }
 
+  @Operation(summary = "Download a user workspace")
```
timcadman, line 399: Should this be SU only?

armadillo/src/main/java/org/molgenis/armadillo/controller/DataController.java
```diff
@@ -490,6 +505,33 @@ public void saveUserWorkspace(
         .get();
   }
 
+  @Operation(summary = "Upload a workspace for a specific user")
```
timcadman, line 514: Should this be SU only?

ui/src/components/DataUpload.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div>
+    <FileUpload
+      ref="fileUpload"
+      :uniqueClass="uniqueClass"
+      :triggerUpload="triggerUpload"
+      :uploadFileMethod="uploadDataFile"
+      :isUploadingFile="isUploadingFile"
```
timcadman: Claude: binds :isUploadingFile, but FileUpload declares that as data, not a prop — the binding does nothing, and DataUpload's own isUploadingFile is set true and never reset.

ui/src/components/FileUpload.vue
timcadman, line 1: Claude: FileUpload still declares uploadCsvAsParquet, typeRows and uploadDone after the CSV UI moved to DataUpload — dead state.

timcadman (changes requested):

timcadman (approved):

PR #1046, opened 2026-07-09
 Build: Release test Tulip-Donkey

DickPostma:
how to test:
- Check ci release test

todo:
- [ ] added tests


timcadman (changes requested):
7 failing tests on the release test. The DSI ones, I think, are related to having the latest version of DSI and DSMolgenisArmadillo. The profile ones, I think, are because it is attempting to connect to a default profile which it can't find.

PR #1048, opened 2026-07-14
Build/update image

DickPostma:
Updated the dockerfile and build image and deprecated devtools functions


DickPostma:
Wrong PR.. create a new

PR #1076, opened 2026-08-28
Modify application.yml for profile correction

RubenOPS:


StuartWheater:
The "molgenis-rock-base" profile also contains "dsTidyverse" package

marikaris (approved):

PR #1080, opened 2026-09-15
docs: cas infrastructure

DickPostma:
how to test:

todo:
- [ ] added tests


docs/pages/cas.md
```diff
@@ -0,0 +1,391 @@
...
+
+    P1 --> K8S[Azure Kubernetes Cluster]
+    P2 --> K8S
+    P3 --> K8S
+
+    P1 --> DS[DataSHIELD]
+    P2 --> DS
+    P3 --> DS
+
+    Auth --> ARM[Armadillo Servers]
```
timcadman, line 43: In the map, should Armadillo and DataSHIELD be combined, so it's clear that Armadillo is accessed via CAS? so you have one box Armadillo/DataSHIELD, and this received an arrow from Keycloak and then the profiles?

timcadman (changes requested):
Looks good - just had that one comment and will be good to hear how this matches Stuart's needs too

